PR #1, opened 2020-11-12
Spencer wip

sglyon:


can_tools/scrapers/__init__.py
```diff
@@ -1,10 +1,6 @@
-from .base import CMU, DatasetBaseNeedsDate, DatasetBaseNoDate, InsertWithTempTable
-
+from .base import (CMU, DatasetBaseNeedsDate, DatasetBaseNoDate,
```
ghop02, line 1: 👋  just saw this PR and dropping in with a couple of style questions haha.  Thoughts on using absolute path names? the [google style guide](https://google.github.io/styleguide/pyguide.html#23-packages) recommends them and I think it helps reorganization.  (for instance if you move a file don't have to change all of it's imports)

ghop02: Also if we want to go that direction, I'm happy to go through and change all of the existing imports myself

can_tools/scrapers/official/IL/__init__.py
```diff
@@ -0,0 +1,123 @@
+import pandas as pd
+import requests
+import us
+
+from ... import DatasetBaseNoDate
+from ..base import CMU, CountyData
+
+
+class IllinoisHistorical(DatasetBaseNoDate, CountyData):
```
ghop02, line 9: My other initial thought on first glance through - how would you feel about having separate files for each scraper (that aren't in __init__.py).

I think there's an opportunity to make the modules more specific to the source so it's clear what scrapers we have for a given state.

for example:
```
IL/
  __init__.py
  illinois_dph_scraper.py
```

requirements.txt
```diff
@@ -13,3 +13,7 @@ us
 geopandas
 pytest
 black
+xlrd
+isort
+importmagic
+epc
```
ghop02, line 19: How would you feel about pinning the versions we're using? We may have slightly different python environment philosophies (which is totally fine).  Also, definitely doesn't have to happen (or maybe we could provide for both ways) but we don't use conda at all in our other two repos.  Do you think it's worth keeping repo setup the same across all repos?

ghop02 (commented):
Feel free to ignore any of my comments, or pass any of that work the comments entail over to me, just half acquainting myself with the code half raising some patterns we have in our other repos and seeing if they're worth applying here.

Also happy to chat about any of them!

requirements.txt
sglyon: Totally should -- this requirements.txt is a bit "wip" as I add more things that I use in my python env. 

We'll update this in another issue!

can_tools/scrapers/__init__.py
sglyon: Happy to use absolute path names.

I think I've seen different opinions on this, but overall I find absolute imports to be easier also

can_tools/scrapers/official/IL/__init__.py
sglyon: Also ok with that.

Most of the states only had one scraper, so I got lazy and instead of making a new file for a scraper and then bringing it into the `__init__.py` I just put code directly in `__init__.py`

requirements.txt
sglyon: Another comment -- I should separate my personal preferneces/setup/dependencies for working on this project from what is absolutely necessary to run the scrapers.

Will do when I resolve the issue in #3

can_tools/scrapers/__init__.py
cc7768: I actually prefer relative imports -- I admit that I'm not sure exactly why I prefer them though...

Happy to do absolute imports if the two of you like it better (especially if @sglyon or @ghop02 does the work to update everything we've written so far :laughing: )

can_tools/scrapers/official/IL/__init__.py
cc7768: This is fine with me too.

can_tools/scrapers/__init__.py
sglyon: moved to absolute.

sglyon:
Thanks @cc7768 and @ghop02 

These changes are incorporated over in #2 , so I'll close this now

PR #2, opened 2020-11-12
Cleanup

sglyon:


can_tools/scrapers/base.py
```diff
@@ -51,37 +51,32 @@ def extract_CMU(
...
-            **{
-                c: foo.map(lambda x: cmu[x].__getattribute__(c))
-                for c in columns
-            }
+            **{c: foo.map(lambda x: cmu[x].__getattribute__(c)) for c in columns}
         )
 
     @abstractmethod
-    def put(self, conn, df):
+    def put(self, conn: sa.engine.base.Engine, df: pd.DataFrame):
```
cc7768, line 58: Am I wrong that most of our `put` methods take a string rather than an engine?

can_tools/scrapers/base.py
```diff
@@ -95,41 +90,61 @@ def quit_early(self, date: pd.Timestamp) -> bool:
...
 
     return f"""
-    INSERT INTO data.{t_home} {cols}
+    INSERT INTO {s_home}.{t_home} {cols}
     SELECT {colnames} from {t_temp}
     ON CONFLICT {pk} DO NOTHING;
     """
 
 
+def _ensure_covid_demographics(connstr: str, df: pd.DataFrame):
```
cc7768, line 107: Did we decide to get rid of this method?

can_tools/scrapers/base.py
```diff
@@ -140,4 +155,7 @@ def put(self, connstr: str, df=None):
             msg = "field `pk` must be set on subclass of OnConflictNothingBase"
             raise ValueError(msg)
 
+        if self.data_type == "covid":
```
cc7768, line 158: Do we want to use the `_ensure_covid_demograhpics` method?

I continue to oppose the idea of inserting uninspected meta information into the database, but I'm open to some push back here.

cc7768 (approved):

can_tools/scrapers/base.py
sglyon: Yes, will remove soon. Good catch

can_tools/scrapers/base.py
sglyon: Nope, we shouldn't use it

can_tools/scrapers/base.py
sglyon: good catch, fixed

can_tools/scrapers/base.py
```diff
@@ -54,84 +73,202 @@ def extract_CMU(
...
+def build_on_conflict_do_nothing_query(
+    columns: List[str],
+    dest_table: str,
+    temp_table: str,
+    unique_constraint: str,
+    dest_schema: str = "data",
 ) -> str:
-    colnames = ", ".join(list(df))
+    """
+    Construct sql query to insert data originally from `df` into `s_home.t_home` via the
```
cc7768, line 201:
```suggestion
    Construct sql query to insert data originally from `df` into `dest_schema.dest_table` via the
```

can_tools/scrapers/official/CA/ca_state.py
```diff
@@ -5,20 +5,39 @@
 import requests
 import us
 
-from ... import DatasetBaseNoDate
-from ..base import CMU, StateDashboard
+from can_tools.scrapers.base import DatasetBaseNoDate, CMU
+from can_tools.scrapers.official.base import StateDashboard
 
 
 class OpenDataCali(StateDashboard, ABC):
```
cc7768, line 12: If we end up writing other scrapers for California, we may end up moving this to a file like `CA/common.py` or something like that -- Nothing to do for now, but just a comment that this may be useful in other contexts.

cc7768 (approved):

cc7768 (approved):

PR #10, opened 2020-12-02
Revive early ports

sglyon:
Recursively find DatasetBase subclasses so no need to inherit everywhere

also:

- make CDC requests 5 at a time
- Split CA scraper
- Split + revive TX scrapers


cc7768 (approved):
This all looks good. Merge away once tests pass.

PR #11, opened 2020-12-04
Create md_state.py

ChrisBremer:
Base python code accessing MD ArcGIS.  Modeled after the Florida scraper, passed local validation on my machine


ChrisBremer:
Ah, forgot about the docker validation on my local machine.  I've got the docker desktop application but was having some trouble getting the localized db up and running with the run command in the training docs. I'm not super familiar with docker so if anyone would be willing to give me some pointers, I'd really appreciate it!

PR #15, opened 2020-12-08
Sqlalchemy

sglyon:
OK @cc7768  here is the sqlalchemy rewrite

Here are a few notes:

- Ran black and isort on the whole repo. Sorry, this created noise
- I got rid of the location_types table and just put the string into the locations table under location_type
- I added tables `temp_official_has_location` and `temp_official_no_location`. 
    - These each have the columns a normalized scraper should return AND a column `insert_op`. 
    - These are the "holding" tables that we put or normalized dataframes into, then we INSERT FROM SELECT from the temp to land in covid_official. 
    - The `insert_op` will be generated by the put method on insert time and will allow the put method to cleanup and delete rows from the temp tables once the INSERT FROM SELECT is done
- You can create a dev db with sqlite using
```python
from can_tools import create_dev_engine
engine, Session = create_dev_engine()
```
This will create a new sqlalchemy engine and sessionmaker class using in-memory sqlite and load up all the tables meta rows.
- The `put` family of methods now accept a sqlalchemy engine instead of a connection string
- The `_put_exec` method just needs an engine and DataFrame now -- no need to pass table name or pk
- The hand written sql is gone
- I added a column `locations.state_fips` to be an integer for state fips code. There is a unique constraint on (location_type, state_fips, name). These three columns are used to join in a normalized DataFrame where `has_location = False`
- The bootstrappedtable rows are now contained in csv files in `can_tools/bootstrap_data`. There is one file per table.

I'm sure there are other things in here that I forgot. If you spot any don't hesitate to ask


cc7768 (approved):
This looks great -- I actually kind of like that these changes move most of the SQL commands to "Python versions" of the same commands.

I don't understand all of the intricacies of `sqlalchemy` quite well enough to follow `models.py`, but it made sense from what I could tell. Let's sync for just a couple mins so that I can make sure I understood the pieces.

PR #18, opened 2020-12-09
WIP: Adding Missouri and updating puppet code.

zweissman:
First real PR so feedback welcome.

The code in mo_state is just a copy from Florida, so ignore it other than the tableau_click_selector_list setting up the list of arguments to be sent to puppet.py to get data from Tableau.

The Missouri dashboard was in ARCGIS https://mophep.maps.arcgis.com/apps/MapSeries/index.html?appid=8e01a5d8d8bd4b4f85add006f9e14a9d
   but has been moved to Tableau https://showmestrong.mo.gov/data-download/

The Excel data that they have is not as detailed as what is available in TableauNeedsClick
https://results.mo.gov/t/COVID19/views/COVID-19DataforDownload/MetricsbyTestDatebyCounty?%3AisGuestRedirectFromVizportal=y&%3Aembed=y

The page is protected by F5 to stop bots from accessing the site. Take a look in the cookies and the notes in can_tools/scrapers/puppet.py for more details.

Also, this Tableau sheet requires more clicks to get the data via the Download button. Updated the code to allow the state code to specify the click stream to follow to get to the data via tableau_click_selector_list.

One more fun challenge. When downloading data from Tableau's Data download option, it opens another tab that shows the first 1000 rows of data. There is a link on that new tab that allows you to download ALL the data. Clicking that opens another tab that will download the data. I was unable to find away to set the Page.setDownloadBehavior for this new page, hence, _special_tableau_handler code. This / Any special function can be added to the tableau_click_selector_list and that code will be run as part of the click process. In this case, it will get the download link from the new tab and download that code manually via the Requests module.

Lastly, Windows doesn't like the puppeteer / chromium options --single-process, --no-sandbox. I wrapped them in a os.name != 'nt' check to only include them in non-Windows environments.

### DIFF
--- a/can_tools/scrapers/official/MO/mo_state.py
+++ b/can_tools/scrapers/official/MO/mo_state.py
@@ -1,0 +1,227 @@
+from abc import ABC
+
+import pandas as pd
+
+import pyppeteer
+import us
+from typing import List
+
+from can_tools.scrapers.base import CMU
+from can_tools.scrapers.official.base import StateDashboard
+from can_tools.scrapers.puppet import TableauNeedsClick
+
+
+class MissouriBase(
+    StateDashboard,
+    TableauNeedsClick,
+    ABC,
+):
+    source = "https://results.mo.gov/t/COVID19/views/COVID-19DataforDownload/{tab}?:isGuestRedirectFromVizportal=y&:embed=y"
+    has_location = False
+    location_type = "county"
+    state_fips = int(us.states.lookup("Missouri").fips)
+
+    out_cols = [
+        "dt",
+        "vintage",
+        "location_name",
+        "category",
+        "measurement",
+        "unit",
+        "age",
+        "race",
+        "sex",
+        "value",
+    ]
+
+    #def clean_desoto(self, df: pd.DataFrame):
+    #    df.loc[df["location_name"] == "Desoto", "location_name"] = "DeSoto"
+
+async def _special_tableau_handler(page, tmpdirname):
+    # Get a reference to the page that was just opened
+    pages = await page.browser.pages()
+    popup = pages[-1]
+
+    import requests
+    import os
+    await popup.waitForSelector('.csvLink_summary')
+    url = await popup.querySelectorEval('.csvLink_summary', 'node => node.href')
+    data = requests.get(url)
+    open(os.path.join(tmpdirname, 'Metrics_by_Test_Date_by_County_data.csv'), 'wb').write(data.content)
+
+
+class MissouriState(MissouriBase):
+    """
+    Fetch details about overall Missouri state testing by county
+    """
+    headless = True
+    url = MissouriBase.source.format(tab="MetricsbyTestDatebyCounty")
+
+# TODO0: ZW: The Data tab below is better... I think. Remove this once confirmed which data to use
+#    # Download Crosstab from Tableau
+#    tableau_click_selector_list: List[str] = [
+#        "@data-tb-test-id = 'DownloadCrosstab-Button'",
+#        "@data-tb-test-id = 'crosstab-options-dialog-radio-csv-Label'",
+#        "@data-tb-test-id = 'export-crosstab-export-Button'"
+#    ]
+
+    # Download Data from Tableau
+    tableau_click_selector_list: List[str] = [
+        "@data-tb-test-id = 'DownloadData-Button'",
+        _special_tableau_handler
+    ]
+
+    async def _pre_click(self, page: pyppeteer.page.Page):
+        # TODO0: ZW: This page is slow to load. Adding wait. 
+        #  Should probably find a better selector to use to determine that the page has loaded
+        import asyncio
+        await asyncio.sleep(5)
+        await page.waitForSelector("#main-content")
+
+    def _clean_cols(self, df: pd.DataFrame, bads: list, str_cols: list) -> pd.DataFrame:
+        for col in list(df):
+            for bad in bads:
+                df[col] = df[col].astype(str).str.replace(bad, "")
+            if col not in str_cols:
+                df[col] = pd.to_numeric(df[col])
+
+        return df
+
+    def normalize(self, df: pd.DataFrame) -> pd.DataFrame:
+        bads = [r",", r"%", "nan"]
+        str_cols = ["County", "FileNumber", "ProviderName"]
+        df = self._clean_cols(df, bads, str_cols)
+        df["location_name"] = df["County"].str.title()
+
+        # Rename appropriate columns
+        crename = {
+            "Bed Census": CMU(
+                category="hospital_beds_in_use", measurement="current", unit="beds"
+            ),
+            "Available": CMU(
+                category="hospital_beds_available", measurement="current", unit="beds"
+            ),
+            "Total Staffed Bed Capacity": CMU(
+                category="hospital_beds_capacity", measurement="current", unit="beds"
+            ),
+        }
+
+        # Drop grand total and melt
+        out = (
+            df.query("location_name != 'Grand Total'")
+            .melt(id_vars=["location_name"], value_vars=crename.keys())
+            .dropna()
+        )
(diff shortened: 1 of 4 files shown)


can_tools/scrapers/official/MO/mo_state.py
```diff
@@ -0,0 +1,233 @@
...
+    # def clean_desoto(self, df: pd.DataFrame):
+    #    df.loc[df["location_name"] == "Desoto", "location_name"] = "DeSoto"
+
+
+async def _special_tableau_handler(page, tmpdirname):
+    # Get a reference to the page that was just opened
+    pages = await page.browser.pages()
+    popup = pages[-1]
+
+    import requests
```
sglyon, line 46: Style nitpick -- let's move imports to top of file

can_tools/scrapers/official/MO/mo_state.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # Download Data from Tableau
+    tableau_click_selector_list: List[str] = [
+        "@data-tb-test-id = 'DownloadData-Button'",
+        _special_tableau_handler,
+    ]
+
+    async def _pre_click(self, page: pyppeteer.page.Page):
+        # TODO0: ZW: This page is slow to load. Adding wait.
+        #  Should probably find a better selector to use to determine that the page has loaded
```
sglyon, line 81: It looks like you wait for `#main-content` does that remove the need for sleeping for 5?

can_tools/scrapers/official/MO/mo_state.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # Extract category information and add other context
+        out = self.extract_CMU(out, crename)
+        out["dt"] = self._retrieve_dt("US/Eastern")
+        out["vintage"] = self._retrieve_vintage()
+        # self.clean_desoto(out)
+        return out.loc[:, self.out_cols]
+
+
+# class MissouriICUUsage(MissouriHospitalUsage):
```
sglyon, line 132: What is the status of these commented out scrapers?

can_tools/scrapers/puppet.py
```diff
@@ -42,20 +45,34 @@ def __init__(self, headless: bool = True):
...
         )
         page = await self.browser.newPage()
         await page.setUserAgent(CHROME_AGENT)
         await page.setViewport(DEFAULT_VIEWPORT)
+
+        # Required if site is using F5 Big-IP ASM.
+        # You can tell by looking at the cookies and if you see one called
+        # TSxxxxxxxxxx and/or BIGipServerSDC_xxxxxxx,
+        # https://stackoverflow.com/questions/52330611/anyone-know-what-a-ts-cookie-is-and-what-kind-of-data-its-for
+        await page.setBypassCSP(True)
```
sglyon, line 74: Thank you for finding this!

:100:

can_tools/scrapers/puppet.py
```diff
@@ -144,9 +166,16 @@ async def _post_download(self, tmpdirname: str) -> pd.DataFrame:
         fns = glob.glob(os.path.join(tmpdirname, "*.csv"))
         assert len(fns) == 1
         with open(fns[0], "rb") as f:
-            tsv = io.BytesIO(f.read().decode("UTF-16").encode("UTF-8"))
+            try:
+                # Downloaded data from Crosstab needs this
+                tsv = io.BytesIO(f.read().decode("UTF-16").encode("UTF-8"))
+            except Exception as e:
+                # Downloaded data from Data doesn't
```
sglyon, line 173: What's happening here in this try/except?

Is there a way to check if we need to decode/encode so we can handle the case directly?

can_tools/scrapers/puppet.py
```diff
@@ -144,9 +166,16 @@ async def _post_download(self, tmpdirname: str) -> pd.DataFrame:
...
+                f.seek(0)
+                tsv = io.BytesIO(f.read())
             df = pd.read_csv(tsv, sep="\t")
 
+        # TODO0: ZW: Should we be deleting the tempfile once we are done loading it?
```
sglyon, line 178: I believe it will be automatically cleaned up when the context manager's scope ends in the `_get` method

can_tools/scrapers/puppet.py
```diff
@@ -174,14 +203,23 @@ async def _get(self) -> pd.DataFrame:
...
+                        await selector.click()
+                        await asyncio.sleep(2)
+
+                # Wait for download to finish. 10 cycles of 2 seconds is arbitrary
+                for _ in range(10):
+                    if len(os.listdir(tmpdirname)) > 0:
+                        break
+                    await asyncio.sleep(2)
+                else:
+                    # TODO0: ZW: Throw / report error?
```
sglyon, line 221: Yes, let's throw here as something unexpected happened

sglyon (changes requested):
Wow @zweissman you unfortunately got assigned a difficult one!

Thank you for working on this. Your changes to the pyppeteer class are very welcome. 

I had a few comments/suggestions, but overall I think this is close to ready to merge

can_tools/scrapers/official/MO/mo_state.py
zweissman, line 46: Sorry about that. Agreed. This code is no where near ready, it is just calling code for the puppet. Will move them up.

can_tools/scrapers/official/MO/mo_state.py
zweissman: It is going to have to be one or the other. That's the TODO, get back in there and figure out if I can find a css selector that will indicate the page, and more importantly, the data is loaded. There looks like there is a Tableau spinning arrow that pops up when the data is loading and goes away when it is ready. I am going to see if I can trigger off that.

can_tools/scrapers/official/MO/mo_state.py
zweissman: They are copy/paste of Florida. They are just renamed so there is no naming conflicts. This page is just a calling point for the puppet changes.

can_tools/scrapers/puppet.py
zweissman: I was not able to find a way to do it and try and except is the pythonic way to do deal with this kind of uncertainty. If you have another suggestion, I am more than happy to change.

can_tools/scrapers/puppet.py
zweissman: I do not believe this is true, but I will test. I had a ton of tmp folders floating around while testing, but they may have all been failed attempts that never got the chance to cleanup automatically.

can_tools/scrapers/puppet.py
sglyon: That's ok for me!

sglyon:
Hey @zweissman let me know when you'd like me to take another look at this.

Thanks!

can_tools/scrapers/puppet.py
zweissman: @sglyon: you are correct, the file and folders are automatically cleaned up when the context manager's scope ends. Assuming that it does so successfully. If there are any errors or you stop debugging in the middle of it, it sticks around and that is what I was seeing. Removing my comment.

can_tools/scrapers/puppet.py
zweissman: Added assert

sglyon:
ping @zweissman -- is this ready for another review?

sglyon:
Hey @zweissman I took a closer look at the Missouri dashboard and I think we can get away with using their downloadable files for now

They have county level data at https://results.mo.gov/t/COVID19/views/COVID-19DataforDownload/MetricsbyCounty.csv

They provide info on what the testing metrics are in their FAQ here https://showmestrong.mo.gov/data/public-health/data-methodology-and-documentation/

Would you be willing to work on getting this csv into the system? Perhaps it is best to create a new branch/PR for that csv in case we want to revisit the tableau work you've done here later?

PR #6, opened 2020-11-26
HHS Hospital Data (state level)

cc7768:
Happy Thanksgiving @mikelehen -- This scraper should be able to download the state-level hospital data we've talked about.

FYI: @sglyon @ghop02 @TomGoBravo

PR #20, opened 2020-12-10
HHS facility level data

cc7768:
This adds a scraper for the HHS facility-level hospital data

I haven't checked carefully how this lines up with what counties, but seemed to spot check alright against 1 or 2.

The hospitals in Alaska don't seem to have updated their fips codes in 30-40 years... There's some mapping of old fips codes to where they would be now which was a "real treat"

Let's not merge till we've more carefully checked the data etc..., but it's ready for review

(This PR builds on #6 ).

Attn: @mikelehen @sglyon


mikelehen:
❤️

can_tools/scrapers/official/federal/HHS/common.py
```diff
@@ -0,0 +1,26 @@
+import requests
+
+from typing import Dict
+
+import pandas as pd
+
+from can_tools.scrapers.official.base import FederalDashboard
+
+
+class HHSDataset(FederalDashboard):
```
sglyon, line 10: Let's make this inherit from abc.ABC so it is clear it shouldn't be instantiated or run through tests.

can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+        # Set all missing values (-999999) to nan for all numeric columns
+        numeric_cols = list(df.select_dtypes("number"))
+        df.loc[:, numeric_cols] = df.loc[:, numeric_cols].where(lambda x: x > 0, np.nan)
+
+        # Create new columns that we need
+        df["inpatient_beds_used_covid_7_day_avg"] = df.eval(
+            "total_adult_patients_hospitalized_confirmed_covid_7_day_avg + "
+            "total_pediatric_patients_hospitalized_confirmed_covid_7_day_avg"
```
sglyon, line 77: Do other source  include pediatric in "inpatient beds"?

I have a hunch that inpatient is adult and pediatric, if counted, would be reported separately

can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -0,0 +1,151 @@
...
+            "measurement",
+            "unit",
+            "age",
+            "sex",
+            "race",
+        ]
+
+        # TODO: We could do a different groupby and put this into states
+        # or hospital regions
+        out_county = df_long.groupby(identifier)["value"].sum().reset_index()
```
sglyon, line 140: Is the original data at the "observation" level?

sglyon, line 140: Does it make sense to sum all these metrics? It looks like it is summing 7-day averages

sglyon (changes requested):

mikelehen (commented):
I scrolled through and in general this looks great.  Thank you!  I left a couple drive-by comments, but haven't reviewed the whole thing in detail. 😁

mikelehen:
Sidenote: Is it normal for tests to be failing? 😁

can_tools/scrapers/official/federal/HHS/facility.py
cc7768: The original data has one row for each hospital

cc7768: Yes -- As discussed in Slack, the groupby groups everything into county level data and then sums across the hospitals in that county.

can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -0,0 +1,151 @@
...
+from can_tools.scrapers.base import CMU
+from can_tools.scrapers.official.federal.HHS.common import HHSDataset
+
+
+class HHSReportedPatientImpactHospitalCapacityFacility(HHSDataset):
+    has_location = True
+    location_type = "multiple"
+    source = (
+        "https://healthdata.gov/dataset/"
+        "covid-19-reported-patient-impact-and-hospital-capacity-state-timeseries/"
```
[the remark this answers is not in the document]
cc7768: :pray:

can_tools/bootstrap_data/covid_categories.csv
```diff
@@ -49,3 +49,6 @@ test,unspecified_tests_negative
 test,unspecified_tests_positive
 test,unspecified_tests_total
 test,reported_positivity
+hospital,critical_staff_shortage_yes
+hospital,critical_staff_shortage_no
+hospital,critical_staff_shortage_noreport
```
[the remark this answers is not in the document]
cc7768: This is only applicable for the state timeseries data.

It's related to the coverage information, but I think it also tells us a bit more (which is why I chose to include this one and not the other). In addition to giving us an idea of the number of hospitals that actually report, it is telling us how many of them are currently experiencing (or are anticipating) a staffing shortage.

Not sure if we would make use of it, but it seemed at least potentially relevant and was low cost to include.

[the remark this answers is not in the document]
mikelehen: Cool, I'm definitely down to get the critical_staff_shortage data ingested! 🎉  

I was just highlighting that the "_noreport" column is an aspect that we aren't capturing in the rest of this data ingestion (all of the hospital utilization / capacity numbers could be incomplete and we don't know how incomplete they are without capturing the _coverage columns).

can_tools/scrapers/official/federal/HHS/facility.py
cc7768: This is an important question.

I'm going to vote for using the sum because I think they're using the sum of pediatric and adult beds in other places (they report adult inpatient beds in a separate variable).

Here's the data dictionary: https://healthdata.gov/covid-19-reported-patient-impact-and-hospital-capacity-facility-data-dictionary

can_tools/bootstrap_data/covid_categories.csv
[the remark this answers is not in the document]
cc7768: I see what you mean now -- Completely agree that the `_noreport` column is akin to the `_coverage` data (which I've chosen not to collect for now).

I'm going to leave it for now because it doesn't cost much to have it (and saves me 15 seconds it would take to delete the right data :sweat_smile: ). I'll pitch it more optimistically as, "this way we have a way to estimate how many hospitals are reporting", but if we wanted to do something serious with information on the `_coverage` then we would want to bring it in for each relevant variable.

cc7768:
Unfortunately, it's not uncommon for at least one of the headless scrapers to fail -- We retry running them at least 3 times in the production environment but not on the gh testing.

sglyon (approved):

can_tools/bootstrap_data/covid_categories.csv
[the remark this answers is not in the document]
mikelehen: 👍

PR #21, opened 2020-12-10
Wisconsin State Data w/Demographics; Update to Wisconsin Counties Data; Added Ethnicity to Project

MattSidor:
1. Added Wisconsin state-level scraper which includes Demographics. Updated demographic options in `covid_demographics.csv`.
2. Updated Wisconsin county-level scraper with corrected category and new categories I had missed in my earlier PR (https://github.com/covid-projections/can-scrapers/pull/13)
3. Added "ethnicity" as a demographic option to the project. Updated all scrapers and updated infrastructure pieces. Tests passing for all previously-completed scrapers.


sglyon:
@MattSidor this is a lot of work! Thank you for going through this  :clap: 

I will request a review from both Chase and I as there are a number of changes here that we should all see

can_tools/scrapers/official/IL/il_demographics.py
```diff
@@ -64,9 +64,17 @@ def get(self) -> pd.DataFrame:
                     "race",
                 ).assign(county=name)
             )
+            parts.append(
```
cc7768, line 67: @sglyon -- Do I need to review these files? They don't seem to be actively running in airflow.

can_tools/scrapers/official/WI/wi_state.py
```diff
@@ -5,12 +5,11 @@
 from can_tools.scrapers.official.base import ArcGIS
 
 
-class Wisconsin(ArcGIS, DatasetBase):
+class WisconsinCounties(ArcGIS, DatasetBase):
```
cc7768, line 8: @MattSidor @sglyon 

Given that these scrapers are pretty "big", what are our thoughts on moving them into two separate files?

can_tools/scrapers/official/WI/wi_state.py
```diff
@@ -33,23 +32,480 @@ def normalize(self, data):
...
+    """
+
+    ARCGIS_ID = ""  # not applicable on this server
+    has_location = True
+    location_type = "state"
+    state_fips = int(us.states.lookup("Wisconsin").fips)
+    source = "https://www.dhs.wisconsin.gov/covid-19/data.htm"
+
+    # # override query URL since Wisconsin has a self-hosted ArcGIS server with a different URL pattern
+    def arcgis_query_url(
```
cc7768, line 110: @MattSidor @sglyon 

To simplify this and ensure DRY, I created a new file called `common.py` and a `WisconsinArcGIS` class that inherits from `ArcGIS` and then has a single method which is `arcgis_query_url` then both `WisconsinState` and `WisconsinCounty` can inherit from this and not redefine this method. Thoughts?

can_tools/scrapers/official/WI/wi_state.py
```diff
@@ -33,23 +32,480 @@ def normalize(self, data):
...
+    def normalize(self, data):
+        df = self.arcgis_jsons_to_df(data)
+        df.columns = [x.lower() for x in list(df)]
+        df["location"] = self.state_fips
+
+        crename = {
+            # #
+            # # totals across all sexes/ages/races/ethnicities
+            # #
+            "positive": CMU(
```
cc7768, line 129: @MattSidor @sglyon 

I'm pretty sure that I previously voted that we should label this as `pcr_tests_positive, cumulative, unique_people`, but now I'm wondering if this particular combination of variables should be given `cases, cumulative, people`... :thinking:

The nice thing being that then we'd have good information that was explicitly labeled as cases which is really what that combination of variables gives us. Keen to hear other thoughts on this though.

cc7768 (changes requested):
This is awesome -- I'm really excited to start ingesting proper demographic data.

Many thanks @MattSidor for taking on a pretty big PR. I think it looks very good. I left a few comments for review by you and Spencer (and pushed some changes which either of you can ask me to revert if they're unwanted). Let me know what y'all think.

can_tools/bootstrap_data/covid_variables.csv
```diff
@@ -8,6 +8,7 @@ deaths,rolling_average_7_day,people
 hospital_beds_in_use_covid,cumulative,beds
 hospital_beds_in_use_covid,new,beds
 hospital_beds_in_use_covid,current,beds
+hospital_beds_in_use_covid,cumulative,people
```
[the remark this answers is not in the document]
sglyon: @cc7768 is this true?

Didn't we plan to write a function that would try to insert rows and apply an `ON CONFLICT TO NOTHING` clause at the end?

can_tools/scrapers/official/WI/wi_state.py
sglyon: I'd be ok either way.

can_tools/scrapers/official/WI/wi_state.py
sglyon: I'm ok with that. I don't think we necessarily need a common.py file, but I am not opposed to it

can_tools/scrapers/official/WI/wi_state.py
sglyon: I don't think I totally understand your last comment. Can you elaborate?

can_tools/bootstrap_data/covid_variables.csv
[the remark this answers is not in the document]
cc7768: I suspect you asking if it is true is probably a nice way of saying that it isn't true -- I assumed that even that insert method would get upset because of the ordering, but, if not, then let's definitely keep grouping similar variables

can_tools/scrapers/official/WI/wi_state.py
cc7768: K -- I thought we had started doing more separation of scrapers, but, after looking at some of the others, I think I'm okay leaving these together and just defining the `WisconsinArcGIS` class at the top of the file.

can_tools/scrapers/official/WI/wi_state.py
cc7768: See comments on separating the scrapers -- No need to pull this out.

can_tools/scrapers/official/WI/wi_state.py
cc7768: For the demographic data, it seems like Wisconsin is reporting `pcr_tests_positive, cumulative, unique_people` as basically a fill-in for cases. We are getting `pcr_tests_positive, cumulative, unique_people` for each of the age/race/ethnicity/sex groups -- This feels more like they want us to think about these numbers as cases than testing data (they also report deaths for these groups).

My concrete proposal is that rather than store these as `pcr_tests_positive, cumulative, unique_people` that we call these `cases, cumulative, people`.

can_tools/bootstrap_data/covid_variables.csv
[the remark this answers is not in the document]
MattSidor: Okay, I will leave this as-is, but let me know if it causes any issues after merging.

can_tools/bootstrap_data/covid_variables.csv
```diff
@@ -20,6 +21,7 @@ pediatric_icu_beds_in_use_covid,current,beds
 pediatric_icu_beds_available,current,beds
 pediatric_icu_beds_capacity,current,beds
 icu_beds_in_use_covid,current,beds
+icu_beds_in_use_covid,cumulative,people
```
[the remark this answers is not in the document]
MattSidor: Leaving as-is for now based on @sglyon's comment.

sglyon:
@MattSidor please let me know when this is no longer in draft status and you are ready for a review.

Thank you

MattSidor:
> @MattSidor please let me know when this is no longer in draft status and you are ready for a review.
> 
> Thank you

Ready for review but there are still some unanswered questions from Chase that I think he's awaiting your responses to.

can_tools/scrapers/official/IL/il_demographics.py
sglyon: Let's not review them here. I think I wrote them, but didn't activate them b/c we weren't decided on how to handle demographics. Leaving them as is is great

can_tools/scrapers/official/WI/wi_state.py
sglyon: sounds good to me. Thanks!

sglyon:
I see thanks @MattSidor 

I agree with the proposal from @cc7768  and then think this is ready for merge.

Thanks for the great work here!

can_tools/scrapers/official/WI/wi_state.py
MattSidor: Just to clarify @cc7768, do you want `pcr_tests_positive` changed to `cases` just for the demographic groups, or for all of the data in Wisconsin?

cc7768: I want to use `cases` instead of `pcr_tests_positive` only when we see the combination `pcr_tests_positive, cumulative, unique_people`

MattSidor: Okay, will do. 👍

MattSidor:
@sglyon @cc7768 I think that all loose ends are tied up on this PR and it's ready to merge now.

cc7768:
Will take a look and merge tonight.

Many thanks @MattSidor!

cc7768 (approved):

PR #22, opened 2020-12-13
Convert official state of Tennessee scraper to new format

jrybacek:
Conversion of current state TN scraper `https://github.com/valorumdata/cmdc-tools/blob/master/src/cmdc_tools/datasets/official/TN/data.py` to new format.


can_tools/scrapers/official/TN/tn_state.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # Rename columns
+        df = df.rename(columns={"DATE": "dt"})
+
+        # Create dictionary for columns to map
+        crename = {
+            "TOTAL_CONFIRMED": CMU(
+                category="cases", measurement="cumulative", unit="people"
+            ),
+            "TOTAL_PROBABLE": CMU(category="cases", measurement="new", unit="people"),
```
sglyon, line 44: I believe the interpretation of new is "was reported since yesterday"

If so, I don't think that this is categorized correctly based on name `TOTAL_PROBABLE`

Also FWIW I believe we decided to stop tracking probable cases because of how differently this is reported across states or counties.

@cc7768 please fact check me here

can_tools/scrapers/official/TN/tn_state.py
```diff
@@ -0,0 +1,208 @@
...
+            "TOTAL_DEATHS": CMU(
+                category="deaths", measurement="cumulative", unit="people"
+            ),
+            "NEW_DEATHS": CMU(category="deaths", measurement="new", unit="people"),
+            "POS_TESTS": CMU(
```
sglyon, line 49: Is there any documentation from TN that can help us figure out what type of test it is?

can_tools/scrapers/official/TN/tn_state.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # Rename columns
+        df = df.rename(columns={"DATE": "dt", "COUNTY": "location_name"})
+
+        # Create dictionary for columns to map
+        crename = {
+            "TOTAL_CONFIRMED": CMU(
+                category="cases", measurement="cumulative", unit="people"
+            ),
+            "TOTAL_PROBABLE": CMU(category="cases", measurement="new", unit="people"),
```
sglyon, line 137: See comment above about CMU here

can_tools/scrapers/official/TN/tn_state.py
```diff
@@ -0,0 +1,208 @@
...
+            "POS_TESTS": CMU(
+                category="unspecified_tests_positive",
```
sglyon, line 139: see comment above about finding type of test

can_tools/scrapers/official/TN/tn_state.py
```diff
@@ -0,0 +1,208 @@
+import pandas as pd
+import requests
+import us
+
+from can_tools.scrapers.base import CMU
+from can_tools.scrapers.official.base import CountyDashboard
+from can_tools.scrapers.official.base import StateDashboard
+
+
+class TennesseeCasesDeathsTests(CMU, StateDashboard):
```
sglyon, line 10: I think I would vote that we name this `TennesseeState` and the other class `TennesseeCounty`

Happy to hear alternative viewpoints

can_tools/scrapers/official/TN/tn_state.py
jrybacek: You are correct, I think I was using the old columns and finding their best match before.  I think a more relevant column would be the "NEW_CONFIRMED" column.  I'll update.

can_tools/scrapers/official/TN/tn_state.py
jrybacek: Thanks for asking me to investigate a bit more.  I've found these answers, I think we can safely call this a PCR test.  https://www.tn.gov/content/dam/tn/health/documents/cedep/novel-coronavirus/Regional_Rate_Map_FAQ.pdf
https://www.tn.gov/health/cedep/ncov/data/data-faqs.html (under What is a confirmed case?)  I'll update accordingly.

can_tools/scrapers/official/TN/tn_state.py
jrybacek: Updated code as well here.

can_tools/scrapers/official/TN/tn_state.py
jrybacek: Sounds great, I was thinking about that after I had submitted the pull request.  I'll split them up.  Should we split the `tn_state.py` file into multiple files as well?

can_tools/scrapers/official/TN/tn_state.py
jrybacek: Updated code as well here.

can_tools/scrapers/official/TN/tn_state.py
sglyon: Thanks for the thoughts here. 

I'm excited to see all the data come in!

I think that for each of the datasets in your list above, the second class name listed makes sense to me.

Are you ok with those names?

jrybacek: Yes, I'll get the classes renamed now.

sglyon:
This looks good to me!

Ready to merge?

sglyon (approved):

sglyon:
:partying_face:

PR #61, opened 2020-12-15
A few schema changes:

sglyon:
1. Add `provider_id` to `covid_observations` Primary Key
2. Remove `covid_official` and `covid_usafacts` tables. We only have `covid_observations` now and we can differentiate data from other sources via the `provider_id` column
3. Added an `id` column to the locations table. This is  a String that matches the `locationId` column from the CAN api
4. dropped `location` and `location type` from `covid_observations` in favor of new `Location.id` column
5. Removed `pk` attribute from DatasetBase -- this is now inferred by sqlalchemy for us
6. Changed from `table_name: str` to `table: Type[Base]` on DatasetBase

End result is that:

1. we only have one main covid observations table. The provider is part of the primary key so we will have potentially multiple observations for each (dt, location, variable, demographic), one for each provider. This should enable the features of querying/filtering based on a specific provider (e.g. all data from state dashboards, all data from cdc, etc.)
2. We now have the location_id column that lines up with what CAN uses.

cc @ghop02 @TomGoBravo @mikelehen

### DIFF
--- a/can_tools/bootstrap_data/locations.csv
+++ b/can_tools/bootstrap_data/locations.csv
@@ -1,3290 +1,3290 @@
-"location","state_fips","state","name","area","latitude","longitude","fullname","location_type"
-54,54,"54","West Virginia",24041.943,38.647285,-80.618324,"West Virginia","state"
-12,12,"12","Florida",53649.703,28.45743,-82.40915,"Florida","state"
-17,17,"17","Illinois",55515.605,40.102875,-89.15261,"Illinois","state"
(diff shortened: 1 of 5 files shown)


mikelehen:
👍 

To be clear, what's the relation between this new `provider_id` and the existing `provider` column we're getting in the parquet file we're consuming?

sglyon:
the id is the auto-incrementing integer we use internally. 

The provider column is still what shows up in the parquet file

PR #68, opened 2021-01-04
Created Connecticut Scraper

blew2015:
Created Connecticut scraper which pulls Covid data at the county and state wide levels.


can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,198 @@
...
+
+class CTCountyDeathHospitalizations(SODA, DatasetBase):
+    source = "https://data.ct.gov/bfnu-rgqt.json"
+    state_fips = int(us.states.lookup("Connecticut").fips)
+    baseurl = "https://data.ct.gov/"
+    has_location = False
+    location_type = "county"
+    cdh_id = "bfnu-rgqt"
+    state_fips = int(us.states.lookup("Connecticut").fips)
+    source = "https://data.ct.gov/resource/bfnu-rgqt.json"
```
sglyon, line 19: Looks like `source` was set twice. I think this second one is correct

If so, let's delete the first one

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,198 @@
...
+            "deaths_total": CMU(
+                category="deaths", measurement="cumulative", unit="people"
+            ),
+            "hospital_beds_in_use_covid_total": CMU(
+                category="hospital_beds_in_use_covid",
+                measurement="current",
+                unit="beds",
+            ),
+        }
+        for c in [
```
sglyon, line 55: This `for` loop isn't necessary.

When you call `cdh.melt` below and set `value_vars=crename.keys()`, pandas will ignore all columns that aren't in crename.keys() (or the id_vars)

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,198 @@
...
+            "positive_tests_pcr_total": CMU(
+                category="pcr_tests_positive", measurement="new", unit="specimens"
+            ),
+            "negative_tests_pcr_total": CMU(
+                category="pcr_tests_negative",
+                measurement="new",
+                unit="specimens",
+            ),
+        }
+        for c in [
```
sglyon, line 129: This for loop can also be deleted (see above)

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,198 @@
...
+    source = "https://data.ct.gov/resource/qfkt-uahj.json"
+
+    def fetch(self):
+        tests = self.get_dataset(self.test_id)
+
+        return tests
+
+    def normalize(self, data: Any) -> pd.DataFrame:
+        # Get raw dataframe
+        tests_rename = {
```
sglyon, line 85: It looks like you rename the columns here and then use the updated column names in `crename` below

Would it be possible to use the originanl column names as keys to `crename`? That save some typing!

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,198 @@
...
+    source = "https://data.ct.gov/resource/bfnu-rgqt.json"
+
+    def fetch(self):
+        cdh = self.get_dataset(self.cdh_id)
+
+        return cdh
+
+    def normalize(self, data) -> pd.DataFrame:
+        # fips_df = pd.DataFrame.from_dict(self.state_fips)
+        cdh_rename = {
```
sglyon, line 28: See comment below about renaming columns AND using custom names as keys of `crename` later

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,198 @@
...
+                measurement="cumulative",
+                unit="people",
+            ),
+            "hospitalizedcases": CMU(
+                category="hospital_beds_in_use_covid",
+                measurement="cumulative",
+                unit="beds",
+            ),
+        }
+        for c in [
```
sglyon, line 189: Also not necessary

sglyon (changes requested):
Overall this is very good! Thank you @blew2015 

It looks like there are a lot of comments, but they are mostly 2 comments repeated once of each of the 3 scrapers.

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def normalize(self, data: Any) -> pd.DataFrame:
+        # Get raw dataframe
+        tests_rename = {"date": "dt", "county": "location_name"}
+        tests = data.rename(columns=tests_rename)
+
+        tests["dt"] = pd.to_datetime(tests["dt"])
+
+        tests = tests.reindex(columns=tests.columns)
+        for c in [c for c in tests_rename.values() if "number" in c]:
```
sglyon, line 77: Same two comments here:

- Looks like the `.reindex` call is reduntant
- Loop iterator seems to be empty

sglyon (changes requested):
Hey @blew2015 thanks for making those changes.

A couple more and I think we are there

can_tools/scrapers/official/CT/ct_state.py
```diff
@@ -0,0 +1,164 @@
...
+
+        return cdh
+
+    def normalize(self, data) -> pd.DataFrame:
+        # fips_df = pd.DataFrame.from_dict(self.state_fips)
+        cdh_rename = {"dateupdated": "dt", "county": "location_name"}
+        cdh = data.rename(columns=cdh_rename)
+        cdh["dt"] = pd.to_datetime(cdh["dt"])
+        cdh = cdh.reindex(columns=cdh.columns)
+        for c in [
```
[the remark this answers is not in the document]
blew2015: I messed up here.  I could have sworn I changed cdh_rename to  cdh.columns. Thats what I planned to do but apparently didn't. Should be fixed in the new commit. This has been pushed to after `cdh.melt` is called and is now called on that `value` column.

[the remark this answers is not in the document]
sglyon: Excellent, thank you

PR #74, opened 2021-01-04
Update HHS Facility

cc7768:
Update the HHS facility code to compute averages using the sum and coverage column:

For a given variable `x`, we used to report the column `x_avg` but now we're using `x_sum / x_coverage` to compute the average. It has been verified that these two values line up when they are both available but the point raised by Brett and Michael is that `x_sum` and `x_coverage` are hidden less frequently...

@sglyon @mikelehen


BrettBoval:
For posterity: We've confirmed this behavior with people at CMU and at HHS. They acknowledge the data leakage and confirmed that it would be continued for the foreseeable future.

can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -72,40 +72,59 @@ def normalize(self, data: str) -> pd.DataFrame:
...
-        )
+        vars_to_compute_sum = {
+            "inpatient_beds_used_covid_7_day_sum": (
+                "total_adult_patients_hospitalized_confirmed_covid_7_day_sum",
+                "total_pediatric_patients_hospitalized_confirmed_covid_7_day_sum",
+            ),
+            "inpatient_beds_used_covid_7_day_coverage": (
+                "total_adult_patients_hospitalized_confirmed_covid_7_day_coverage",
+                "total_pediatric_patients_hospitalized_confirmed_covid_7_day_coverage",
+            ),
```
mikelehen, line 83: Hrm...  I think summing the `sum`s and `coverage`s before computing the `sum / coverage` 7-day average may be problematic since `a/b + c/d` != `(a+c)/(b+d)`?

Not sure how big of an issue this in practice (if the `_coverage` numbers are identical which seems likely then it'll work out), but I think it'd be more correct to compute the `sum/coverage` average for adults and pediatrics separately and then sum them.

mikelehen: Er, actually I think this will definitely be problematic.  Consider the common case where there are lots of adults (e.g. 100) and very few pediatrics (e.g. 0) and they both have 7 days of coverage, your `coverage` will end up being `14` instead of `7` and so you'll basically be dividing the adult 7-day average in half.

cc7768: :facepalm: :face_with_head_bandage: 

Thank you for catching this. Will update accordingly.

sglyon: @cc7768 once this change is ready I will review

mikelehen (approved):
Looks good to me.

cc7768:
@sglyon -- This received some updates. A review when you're ready would be great.

sglyon (approved):

PR #76, opened 2021-01-06
AZ Scraper Addition

ChrisBremer:
AZ Tableau page scraper - passing pytest on local machine.

This scraper currently requires BeautifulSoup4, so I added it to the requirements-dev.txt.  But I'm currently not sure if that will sync to the live/online test.


sglyon:
Thanks @ChrisBremer -- will you add BeautifulSoup4 to the `requirements.txt` instead of the dev one?

The requirements.txt should contain all hard dependencies of the scrapers

The `-dev` version contains just developer tooling to make our editors/IDEs more productive.

ChrisBremer:
Added BS4 to requirements.txt

can_tools/scrapers/official/AZ/AZ_TableauPull.py
```diff
@@ -0,0 +1,280 @@
...
+            else:
+                return None
+
+        # Let's start at the state-level
+        url = "https://tableau.azdhs.gov/views/COVID-19Summary/Overview2"
+        # Initialize main page: grab session ID key, sheet ID key, root directory string
+        r = requests.get(url, params=self.reqParams)
+
+        # Parse the output, return a json so we can build a bootstrap call
+        suppe = BeautifulSoup(r.text, "html.parser")
```
cc7768, line 139: Are we using anything special from `"html.parser"`? If not, I think we can use the `lxml` parser -- `lxml` is already listed as a dependency in the `requirements.txt`

can_tools/scrapers/official/AZ/AZ_TableauPull.py
```diff
@@ -0,0 +1,280 @@
...
+                category="unspecified_tests_total",
+                measurement="cumulative",
+                unit="specimens",
+            ),
+            "Number of tests serology": CMU(
+                category="antibody_tests_total",
+                measurement="cumulative",
+                unit="specimens",
+            ),
+            "CumPosTests": CMU(
```
cc7768, line 239: Is the unspecified due to the fact that `CumPosTests` could be (and probably is) both diagnostics and serology?

can_tools/scrapers/official/AZ/AZ_TableauPull.py
```diff
@@ -0,0 +1,280 @@
...
+                category="antibody_tests_positive",
+                measurement="cumulative",
+                unit="specimens",
+            ),
+        }
+
+        out = (
+            df.melt(id_vars=["location"], value_vars=crename.keys())
+            .assign(
+                dt=self._retrieve_dt("US/Eastern"), vintage=self._retrieve_vintage()
```
cc7768, line 259: Arizona should use ` US/Arizona` as the timezone in `self._retrieve_dt` method (we must not have told you that it should refer to a location's local time)

can_tools/scrapers/official/AZ/AZ_state.py
```diff
@@ -0,0 +1,280 @@
...
+                category="antibody_tests_positive",
+                measurement="cumulative",
+                unit="specimens",
+            ),
+        }
+
+        out = (
+            df.melt(id_vars=["location"], value_vars=crename.keys())
+            .assign(
+                dt=self._retrieve_dt("US/Eastern"), vintage=self._retrieve_vintage()
```
cc7768, line 259: See tz comment on other file.

can_tools/scrapers/official/AZ/AZ_state.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    Fetch county level covid data from Arizona's Tableau dashboard
+    """
+
+    # Initialize
+    source = "https://www.azdhs.gov/preparedness/epidemiology-disease-control/infectious-disease-epidemiology/covid-19/dashboards/index.php"
+    has_location = True
+    location_type = "state"
+    state_fips = 4
+    cntys = [
```
cc7768, line 21: Do we use all of the counties in the `ArizonaStateData` scraper? Or is this just state data? If we do use them, we might want to move them to a parent class that both `ArizonaStateData` and `ArizonaData` inherit from so that we only have to make changes in one place.

can_tools/scrapers/official/AZ/AZ_state.py
```diff
@@ -0,0 +1,280 @@
...
+class ArizonaStateData(StateDashboard):
+    """
+    Fetch county level covid data from Arizona's Tableau dashboard
+    """
+
+    # Initialize
+    source = "https://www.azdhs.gov/preparedness/epidemiology-disease-control/infectious-disease-epidemiology/covid-19/dashboards/index.php"
+    has_location = True
+    location_type = "state"
+    state_fips = 4
```
cc7768, line 20: I know AZ is fips code 4, but could we set the `state_fips` using `int(us.states.lookup("Arizona").fips)`?

We instituted this because we have previously introduced incorrect data by setting them incorrectly (we forgot to change copy-pasted codes etc...)

can_tools/scrapers/official/AZ/AZ_TableauPull.py
```diff
@@ -0,0 +1,280 @@
...
+class ArizonaData(StateDashboard):
+    """
+    Fetch county level covid data from Arizona's Tableau dashboard
+    """
+
+    # Initialize
+    source = "https://www.azdhs.gov/preparedness/epidemiology-disease-control/infectious-disease-epidemiology/covid-19/dashboards/index.php"
+    has_location = True
+    location_type = "county"
+    state_fips = 4
```
cc7768, line 20: I know AZ is fips code 4, but could we set the `state_fips` using `int(us.states.lookup("Arizona").fips)`?

We instituted this because we have previously introduced incorrect data by setting them incorrectly (we forgot to change copy-pasted codes etc...)

can_tools/scrapers/official/AZ/AZ_state.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def normalize(self, data):
+        df = data
+        crename = {
+            "New Cases": CMU(category="cases", measurement="new", unit="people"),
+            "New Deaths": CMU(category="deaths", measurement="new", unit="people"),
+            "New Tested": CMU(
+                category="unspecified_tests_total", measurement="new", unit="specimens"
+            ),
+            "New Tested PCR": CMU(
```
cc7768, line 212: `New Tested PCR` seems like it would be just `pcr_tests_total`, no?

can_tools/scrapers/official/AZ/AZ_TableauPull.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def normalize(self, data):
+        df = data
+        crename = {
+            "New Cases": CMU(category="cases", measurement="new", unit="people"),
+            "New Deaths": CMU(category="deaths", measurement="new", unit="people"),
+            "New Tested": CMU(
+                category="unspecified_tests_total", measurement="new", unit="specimens"
+            ),
+            "New Tested PCR": CMU(
```
cc7768, line 212: `"New Tested PCR"` seems like it should be just `pcr_tests_total`?

cc7768 (changes requested):
Hey @ChrisBremer -- This looks great! Thanks for taking on a tough scraper :clap: 

I've made a few comments (mostly pretty minor). The biggest open question that I want to raise is how much of this code can be put into a parent class so that, as much as possible, we don't repeat ourselves and leave ourselves open to copy-paste bugs.

can_tools/scrapers/official/AZ/AZ_TableauPull.py
ChrisBremer: This is correct - the value is a derivation from "Percent positive" value, which includes PCR, antigen, serology together.

can_tools/scrapers/official/AZ/AZ_TableauPull.py
ChrisBremer: Ah, silly error on my part, I thought maybe this was a reference to when the query is performed from a machine on the east coast.  Thank you!

can_tools/scrapers/official/AZ/AZ_TableauPull.py
ChrisBremer: Can do.  I had generated my own FIPS code list for testing purposes since Tableau will need an all-caps list of county names anyways for the subview generation.  But that list can be generated iteratively

can_tools/scrapers/official/AZ/AZ_TableauPull.py
```diff
@@ -0,0 +1,280 @@
...
+                if fn.get("name") == "Map filter":
+                    urlFltr = (
+                        urllib.parse.unquote(fn.get("linkSpec").get("url"))
+                        .split("?")[1]
+                        .replace("=<Countynm1~na>", "")
+                        .split("&")
+                    )
+            return urlFltr
+
+        def extractData(htmdump, area):  # Extracts data from raw tableau HTML bootstrap
```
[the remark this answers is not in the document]
ChrisBremer: I think it might generalize, but like we saw with California data, different states are using Tableau differently, and that will impact how we form a parent class. 

From what I've seen, we almost want a few different Tableau parent classes:

    -  Mapclick Subview Data Extraction (Arizona-like):  Pull the onclick map subview query function.  Then use the filter expression it to generate per-county subviews. Then scrape data from these subviews.
    
    -  External Inpage Filter (California-like):  The query expression is not linked to a mapclick, but externally on the page. We need to find that filter expression, pipe it to Tableau to generate county-specific views (these are not subviews), and then scrape the data from there.

can_tools/scrapers/official/AZ/AZ_TableauPull.py
ChrisBremer: This is the html.parser that comes along with BS4 (no additional library install needed).  Did you want to move away from BS4 entirely in preference of lxml?  I can make these updates, but thought we already said BS4 was ok to add to the requirements list.

can_tools/scrapers/official/AZ/AZ_TableauPull.py
ChrisBremer: I thought so too, but according to the AZ dashboard, this column includes both PCR and antigen testing [(see the last '***' annotation](https://www.azdhs.gov/preparedness/epidemiology-disease-control/infectious-disease-epidemiology/covid-19/dashboards/index.php)).   The column name is determined from the Tableau-internal name.  I agree it's not super intuitive

ChrisBremer:
If we're concerned about copy-paste errors, would we be able to add a new location type that is both state and county together?

Based on the way I need to call the state page first to grab the county filter expression before I can do the county scraping, I really don't want to separate them into different scripts.  But instead of struggling with location_type handling, I just split them it in two.

Even if I create a parent Tableau class, it will wind up calling the state page first.

can_tools/scrapers/official/AZ/AZ_TableauPull.py
sglyon: I'm ok to have both as dependencies.

While they are quite similar in scope and we *could* use just one, I don't think it is necessary. Both are mature and easy to install so it shouldn't cause any problems for deployment

cc7768: Definitely okay with using BS4 and putting it in the requirements list -- This was more a comment that you can use different underlying parsers behind BS4. The [BS4 docs](https://beautiful-soup-4.readthedocs.io/en/latest/#installing-a-parser) claim that lxml is both flexible and fast but the cost is that it requires an external dependency so I just meant that since we already  have `lxml` as a dependency then do we want to use the `lxml` parser for BS4.

If speed isn't an issue then I wouldn't worry about it.

ChrisBremer: Oh that makes sense! I'll see about using that instead, I thought you had meant as an alternative.

ChrisBremer:
Latest version uploaded with parent class is implemented - decided to keep state and county separate because the county pull takes a long time (~30sec) to run.

can_tools/scrapers/official/AZ/AZ_state.py
ChrisBremer: Same as the other file - internal nomenclature of the tableau page, per AZ's footnotes this is antigen and PCR.

image

PR #78, opened 2021-01-08
Add 6 days to collection_week to indicate the last data point included.

mikelehen:
@ghop02 pointed out that collection_week indicates the _start_ of the week.

Per the example on https://healthdata.gov/dataset/covid-19-reported-patient-impact-and-hospital-capacity-facility:

> For a given entry, the term “collection_week” signifies the start of the period that is aggregated. For example, a “collection_week” of 2020-11-20 means the average/sum/coverage of the elements captured from that given facility starting and including Friday, November 20, 2020, and ending and including reports for Thursday, November 26, 2020.

So I've added 6 days to the collection_week date.

### DIFF
--- a/can_tools/scrapers/official/federal/HHS/facility.py
+++ b/can_tools/scrapers/official/federal/HHS/facility.py
@@ -1,4 +1,5 @@
 from io import StringIO
+from datetime import timedelta
 
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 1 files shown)


cc7768 (approved):

mikelehen:
I just noticed / realized that the old (pre-date-adjusted) data is still in the database, leading to a weird double-reporting 6 days apart in the data, e.g. as seen here in our charts:

image

This also makes our graphs all stair-steppy:
image

Is this easy to fix somehow?

PR #79, opened 2021-01-08
Fixes pandas SettingWithCopyError warnings.

mikelehen:
This fixes two SettingWithCopyError warnings.  Feel free to holler if there's a better solution. I'm a pandas novice.


sglyon:
This is the way to do it!

Thank you

PR #80, opened 2021-01-11
Iowa County-Level Official Scraper (closes #41)

### LINKED ISSUE #41: Iowa

christopherturner:
Good evening!  I've nearly completed the Iowa scraper port from the `cmdc-tools` version and was hoping to begin reviewing my code!

I'm currently running into 2 errors when running `pytest -v -k Iowa .`, having to do with the insertion constraints.  It appears to be a problem with locations.  For context, while the `IowaCasesDeaths` class has county names (or even FIPS codes if we want them), the `IowaHospitals` class refers to custom "regions".  Are there any recommendations on how to proceed?

### DIFF
--- a/can_tools/scrapers/official/IA/__init__.py
+++ b/can_tools/scrapers/official/IA/__init__.py
@@ -1,0 +1,4 @@
+from can_tools.scrapers.official.IA.ia_state import (
+    IowaCasesDeaths,
+    IowaHospitals,
+)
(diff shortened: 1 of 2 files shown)


sglyon:
Great work @christopherturner 

We do have room in the schema for reporting by hospital reagion. It is a location type here: https://github.com/covid-projections/can-scrapers/blob/main/can_tools/bootstrap_data/location_types.csv

We would need to determine how to append the proper rows to the locations.csv file in the bootstrap_data folder

We may need some insight from @TomGoBravo, @ghop02, and @mikelehen on how to set the locationID column for a hospital region

mikelehen:
I don't have too much context on how we did in the past.  I think @ghop02 might.

PR #84, opened 2021-01-12
CDC Testing Data: Add DC support

### LINKED ISSUE #81: CDC Testing Data does not include DC
It looks like the CDC testing data scraper does not include DC. Unsure of why, as DC is included in `us.STATES`

cc7768:
We are using the `us` package v2 which removed DC from the `us.STATES` list (see https://github.com/unitedstates/python-us#200). In `us` v3, DC can be added to the `us.STATES` list by setting an env variable. We can revisit this update if we decide to upgrade.

We create a new variable `ALL_US_STATES_PLUS_DC` and store it in `can_tools/scrapers/base.py`. This variable is then referenced in the `CDCCovidDataTracker.fetch` method. We can update other scrapers in the future as we notice that they use `us.STATES` to iterate rather than `ALL_US_STATES_PLUS_DC`.

Resolves #81

### DIFF
--- a/can_tools/scrapers/base.py
+++ b/can_tools/scrapers/base.py
@@ -1,12 +1,18 @@
 from can_tools.models import Base
+
 import os
 import pickle
+
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Type
 
 import pandas as pd
+import us
 from sqlalchemy.engine.base import Engine
+
+
+ALL_STATES_PLUS_DC = us.STATES + [us.states.DC]
 
 
 class CMU:
--- a/can_tools/scrapers/base.py
+++ b/can_tools/scrapers/base.py
@@ -12,6 +12,10 @@ import us
 from sqlalchemy.engine.base import Engine
 
 
+# `us` v2.0 removed DC from the `us.STATES` list, so we are creating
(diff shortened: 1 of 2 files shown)


can_tools/scrapers/base.py
```diff
@@ -1,14 +1,20 @@
...
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Type
 
 import pandas as pd
+import us
 from sqlalchemy.engine.base import Engine
 
 
+ALL_STATES_PLUS_DC = us.STATES + [us.states.DC]
```
ghop02, line 15: Maybe a comment about us state versioning would be helpful

ghop02 (approved):

can_tools/scrapers/base.py
sglyon: Agreed, @cc7768 let's add a comment (similar to what you put in the PR) so we know what this was for in a couple weeks

sglyon (changes requested):

can_tools/scrapers/base.py
cc7768: Good thought... I often forget why I made changes within hours (not to mention what can be forgotten over weeks/months).

I've updated the PR to reflect this.

PR #88, opened 2021-01-14
CDC: Update vaccine scraper to reflect the new json structure and new data

cc7768:
@ghop02 @sglyon 

Chris noticed that the structure of the json has changed and now includes first doses and total doses. We're implicitly assuming that `dose 1 + dose 2 = total doses` and imputing `dose 2` values.

Additionally, rather than iterate on `us.STATES`, we use the new `ALL_STATES_PLUS_DC` variable

### DIFF
--- a/can_tools/scrapers/official/federal/CDC/cdc_state_vaccines.py
+++ b/can_tools/scrapers/official/federal/CDC/cdc_state_vaccines.py
@@ -36,9 +36,6 @@ class CDCStateVaccine(FederalDashboard):
         df.loc[:, "location"] = df["Location"].map(
             lambda x: int(us.states.lookup(x).fips)
         )
-        df.loc[:, "Administered_Dose2"] = df.eval(
-            "Doses_Administered - Administered_Dose1"
-        )
 
         crename = {
             "Doses_Distributed": CMU(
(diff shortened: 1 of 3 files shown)


can_tools/scrapers/official/federal/CDC/cdc_state_vaccines.py
```diff
@@ -43,20 +43,15 @@ def normalize(self, data):
...
                 unit="people",
             ),
-            "Administered_Moderna": CMU(
-                category="moderna_vaccine_initiated",
-                measurement="cumulative",
-                unit="people",
-            ),
-            "Administered_Pfizer": CMU(
-                category="pfizer_vaccine_initiated",
+            "Doses_Administered": CMU(
```
ghop02, line 51: Hmm my first thought was to include both what you had in the previous commit and this field. This will just require a bit more logic (probably in covid-data-public), but maybe thats the right spot for it?

ghop02: Ah - maybe I'm coming around to it - it's not really a problem here since its just from the cdc, but will require a bit more logic for the state dashboards

cc7768: Yes -- I went back and forth about it a little bit, but I think this fits the goal of "directly reading in the data via the CAN scrapers and processing it in the ingestion/staging steps". I could be convinced otherwise, but if you agree then I'm happy with it :grin: 

When you say it will require "a bit more logic for the state dashboards", are you talking about the fact that some will report dose 1 and dose 2 while others will report dose 1 and total doses? I think you can use `pd.DataFrame.where` (or something close to that) to use `total - dose_1` or `dose_2` based on what's available for each location. I think that's a valid concern though and am happy to chat more about it.

ghop02: 👍

ghop02 (approved):
LGTM!

sglyon (approved):

PR #97, opened 2021-01-21
ENH: PA vaccine county data

cc7768:
This adds a new class (`MicrosoftBIDashboard`) to `can_tools/scrapers/official/base.py` that provides some helper functionality for working with Microsoft Power BI dashboards.

This class is then used to scrape the county vaccine data from PA.


sglyon:
This is awesome!

I'm not sure what black magic you using in the power BI scraper (read that as I didn't review that part  carefully ;) ), but the data checks out!

Good work

PR #102, opened 2021-01-23
Delaware Addition

ChrisBremer:
Delaware was fairly easy - and my third different type of scraper.

Delaware stores their data in easy-to-access csv files, so no need for any ArcGIS or Tableau.


sglyon:
Go @ChrisBremer !

PR #111, opened 2021-02-01
add scraper for vaccine in LA county

sglyon:


cc7768 (commented):
Overall looks great -- Thank you for writing this and making it sensitive to any change they make going pretty forward.

I have a couple questions:

1. Are we sure that the "Doses Administered" number is first doses? Is it not a total of first and second doses?
2. There is less we can do about this one, but it seems that they're inferring the number of people vaccinated by subtracting remaining inventory from the number of doses they were allocated. This seems to be more of a lower bound in practice because many places seem to extract more than 5 doses per vial.

sglyon:
Good points. I actually don't know   about whether total is does1 or dose1+ dose2. Thinking about it more I would imagine it is probably dose1+dose2.

sglyon:
Per the table names I believe the doses number is total doses.

Are we comfortable taking total doses and completed as reported and then computing `initiated = total - 2 * completed`?

cc @cc7768

cc7768:
> Per the table names I believe the doses number is total doses.
> 
> Are we comfortable taking total doses and completed as reported and then computing `initiated = total - 2 * completed`?
> 
> cc @cc7768

I think that is outside of the purview of scraping the data. My understanding is that those types of transformations should happen later in the pipeline.

sglyon:
SGTM I'll rename the column and we can merge

sglyon:
@cc7768 ready for review now

PR #113, opened 2021-02-02
Add scraper for Missouri

sglyon:


can_tools/scrapers/official/MO/mo_vaccine.py
```diff
@@ -0,0 +1,47 @@
...
+import requests
+import us
+
+from can_tools.scrapers.base import CMU
+from can_tools.scrapers.official.base import TableauDashboard
+
+
+class MissouriVaccineCounty(TableauDashboard):
+    has_location = False
+    source = "https://results.mo.gov/t/COVID19/views/VaccinationsDashboard/Vaccinations?%3Aembed=y&%3AshowVizHome=no&%3Ahost_url=https%3A%2F%2Fresults.mo.gov%2F&%3Aembed_code_version=3&%3Atabs=no&%3Atoolbar=no&%3AshowAppBanner=false&%3Adisplay_spinner=no&%3AloadOrderID=0"
```
cc7768, line 11: Do you need the source to be this entire url?

cc7768 (approved):
Data matches the dashboard. This should be good to go.

can_tools/scrapers/official/MO/mo_vaccine.py
sglyon: No, we could try to limit it. I just copy /pasted what was in my browser hoping that when people click it they'll get to same page. I'll check while I do formatting

PR #118, opened 2021-02-04
ENH: ME vaccines

cc7768:


can_tools/scrapers/official/ME/__init__.py
```diff
@@ -0,0 +1 @@
+from .me_vaccines import MaineCountyVaccines
```
sglyon, line 1: We've used global imports everywhere, going to change now

sglyon (changes requested):

sglyon (approved):

can_tools/scrapers/official/ME/__init__.py
cc7768: Ha! Ooops. Thanks for catching this.

PR #120, opened 2021-02-06
fixes for florida county vaccine scraper

smcclure17:
- Updated the way the dataframe is cleaned/scraped to hopefully prevent future errors (if the placement/formatting of the table changes mildly it should no longer break, hopefully).
- Updated the normalize method to be able to handle if the table overflows onto the second page of the PDF

**Notes:**
My solution to fixing the first problem (how to detect when the data in the table actually begins) feels a bit fragile, but I haven't thought of a better solution. It has worked for the past two weeks (that's as far back as I tested) so it seems stable, but any ideas/advice is welcomed.


cc7768 (approved):
Makes sense -- I wish there was a less fragile way to do it, but sometimes you just have to do what works.

Thanks for the fix!

PR #121, opened 2021-02-06
Scrape official state of New Mexico vaccine data

jrybacek:
Public URL: https://cvvaccine.nmhealth.org/public-dashboard.html
REST API: https://cvvaccine.nmhealth.org/api/GetCounties

### DIFF
--- a/can_tools/scrapers/official/NM/nm_vaccine.py
+++ b/can_tools/scrapers/official/NM/nm_vaccine.py
@@ -75,9 +75,9 @@ class NewMexicoVaccineCounty(NewMexicoBase):
                 unit="doses",
             ),
             "dosesAdministered": CMU(
-                category="total_vaccine_doses_administered",
+                category="total_vaccine_initiated",
                 measurement="cumulative",
-                unit="doses",
+                unit="people",
             ),
             "totalShipped": CMU(
                 category="total_vaccine_distributed",
@@ -115,14 +115,6 @@ class NewMexicoVaccineCounty(NewMexicoBase):
 
         # Drop extraneous columns
         out = df.loc[:, cols_to_keep]
-
-        # Drop the information that we won't be keeping track of
-        # loc_not_keep = ["Out of State", "Pending"]
-        # out = out.query("location_name not in @loc_not_keep")
-
-        # Fix incorrectly spelled county names
-        # loc_replacer = {"Dekalb": "DeKalb"}
-        # out = out.replace({"location_name": loc_replacer})
 
         # Convert value columns
         out["value"] = out["value"].astype(int)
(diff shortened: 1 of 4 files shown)


can_tools/scrapers/official/NM/nm_vaccine.py
```diff
@@ -0,0 +1,137 @@
...
+        # Convert value columns
+        out["value"] = out["value"].astype(int)
+
+        # Add rows that don't change
+        out["vintage"] = self._retrieve_vintage()
+
+        return out
+
+    def validate(self, df, df_hist) -> bool:
+        "Software never was perfect and won’t get perfect. But is that a license to create garbage? The missing ingredient is our reluctance to quantify quality. - Boris Beizer"
```
cc7768, line 136: I always enjoy the docs on your validate methods :smile:

cc7768 (approved):
I made a few very small changes but this was excellent and super easy to review!

can_tools/scrapers/official/NM/nm_vaccine.py
```diff
@@ -0,0 +1,137 @@
...
+                category="moderna_vaccine_distributed",
+                measurement="cumulative",
+                unit="doses",
+            ),
+            "pfizerShipped": CMU(
+                category="pfizer_vaccine_distributed",
+                measurement="cumulative",
+                unit="doses",
+            ),
+            "dosesAdministered": CMU(
```
cc7768, line 77: The one thing that was kind of funny is that `dosesAdministered` seems to line up with the number on their dashboard that is used to denote "Individuals Partially Or Fully Vaccinated". I changed this to reflect what seems to be being reported rather than what they have chosen to name it... We should keep an eye on this because they might update it at some point

For example, in Catron (On February 5) the dashboard reports that 666 were partially or fully vaccinated and the `dosesAdministered` is 666.

cc7768:
Going to merge with my changes but when you have a minute (@jrybacek) will you sanity check my change of the variable... Their naming convention is a strange one relative to other dashboards and I just want to get a second pair of eyes to confirm.

can_tools/scrapers/official/NM/nm_vaccine.py
jrybacek: I guess I take `dosesAdministered` to mean X number of needles in arms.  I agree with the CMU (category="total_vaccine_doses_administered", measurement="cumulative", unit="doses").  I'm not sure if that was mine or yours post modification, but it looks good to me based on my current understanding.

cc7768: That was yours prior to modification -- I changed it to (`total_vaccine_initiated`, `cumulative`, `people`) because how how the number associated with that value lined up with values in the dashboard.

I definitely agree that `dosesAdministered` sounds more like (`total_vaccine_doses_administered`, `cumulative`, `doses`) which is why I wanted a second pair of eyes on this change :astonished:

PR #123, opened 2021-02-08
add src

sglyon:
- black fix me vaccine
- add source_url to observations table and api endpoint


can_tools/scrapers/base.py
```diff
@@ -91,7 +92,7 @@ def quit_early(self) -> bool:
 
     def _retrieve_dt(self, tz: str = "US/Eastern") -> pd.Timestamp:
         """Get the current datetime in a specific timezone"""
-        out = pd.Timestamp.utcnow().tz_convert(tz).normalize().tz_localize(None)
+        out = self.execution_dt.tz_convert(tz).normalize().tz_localize(None)
```
cc7768, line 95: I updated this as part of fixing the database. For the scrapers that use the `_retrieve_dt` to get the information, using the current date doesn't make sense for when they are re-run...

Same logic for `_retrieve_vintage`

can_tools/scrapers/__init__.py
```diff
@@ -7,6 +7,9 @@
     CaliforniaTesting,
     CDCCovidDataTracker,
     CDCStateVaccine,
+    CTCountyDeathHospitalizations,
```
cc7768, line 10: For some reason, these scrapers never made it into the `__init__` file :man_shrugging:

can_tools/scrapers/official/base.py
```diff
@@ -132,6 +136,9 @@ def _prep_df(self, df: pd.DataFrame) -> Tuple[pd.DataFrame, str]:
         if "location_type" not in list(to_ins):
             to_ins["location_type"] = self.location_type
 
+        if "source_url" not in list(to_ins):
```
cc7768, line 139: We needed this logic in the `FederalDashboard` as well so I added it.

can_tools/scrapers/usafacts/data.py
```diff
@@ -30,41 +30,44 @@ def fetch(self) -> pd.DataFrame:
         return pd.read_csv(BASEURL + self.filename)
 
     def normalize(self, data: Any) -> pd.DataFrame:
+        # Make lowercase so they can't change capitalization on us
+        data.columns = [c.lower() for c in data.columns]
```
cc7768, line 34: USAFacts changed the capitalization of their columns -- This prevents us from being exposed to a future change in capitalization.

can_tools/scrapers/usafacts/data.py
```diff
@@ -30,41 +30,44 @@ def fetch(self) -> pd.DataFrame:
...
+
         # Load data from site and move dates from column names to
         # a new variable
-        cols = ["countyFIPS", "County Name", "State", "stateFIPS"]
-        cols = cols + [c for c in data.columns if c[-2:] in ("20", "21")]
+        cols = ["countyfips", "county name", "state", "statefips"]
+        cols = cols + [c for c in data.columns if ("2020" in c) or ("2021" in c)]
```
cc7768, line 39: They also changed how they are formatting their date and they're using a proper date format so we can just check for years now.

cc7768 (approved):
@sglyon -- I made a few additional changes as I did the work to fix the database. Please have a look at these (small) changes.

I will probably merge now so that we can get the USAFacts scrapers running again and because the main changes are related to adding the source as a column which I've reviewed and approve of.

PR #133, opened 2021-02-16
ENH: Add GA vaccine scraper

cc7768:


sglyon (changes requested):
I don't think we got definitions quite right.

The data lines up for total and completed, but for the few counties I hand-checked the number for total doses and initiated (1st dose) were the same.

This is not true on the dashboard

Screenshot_20210216_155624


Screenshot_20210216_155648

cc7768:
> I don't think we got definitions quite right.
> 
> The data lines up for total and completed, but for the few counties I hand-checked the number for total doses and initiated (1st dose) were the same.
> 
> This is not true on the dashboard
> 
> Screenshot_20210216_155624
> 
> Screenshot_20210216_155648

This is on purpose -- The first dose column is first dose only so we're summing first dose and second dose to get "at least one dose" which is our `total_vaccines_initiated` column.

In places where first dose + second dose = total the `total_vaccines_initiated` and `total_doses_administered` should line up.

PR #134, opened 2021-02-17
ENH: Add NV vaccine scraper

cc7768:
There is time-series data buried in their database and I could get it out (partially?) but, when I aggregated it, it didn't exactly line up with the numbers from the dashboard...

For now we'll skip the time-series and potentially revisit in the future.


sglyon (commented):
great work!

sglyon (approved):

PR #139, opened 2021-02-22
ny vaccine

sglyon:
- wip ny -- still doesn'st have data we need
- NY vaccine data
- Simplify some tableau classes -- a few are fully declarative now!


cc7768 (approved):
I like the updates to the class!

can_tools/scrapers/official/base.py
```diff
@@ -511,6 +511,38 @@ class TableauDashboard(StateDashboard, ABC):
...
     filterFunctionValue: Optional[str] = None
+    timezone: str
+    data_tableau_table: str
+    location_name_col: str
+    cmus: Dict[str, CMU]
+
+    def fetch(self) -> pd.DataFrame:
+        return self.get_tableau_view()[self.data_tableau_table]
+
+    def __normalize(self, df: pd.DataFrame) -> pd.DataFrame:
```
ghop02, line 522: I would stay away from dunder's if possible ([google style guide](https://google.github.io/styleguide/pyguide.html#3162-naming-conventions) gives a good explanation) - would `_normalize` do the trick?

PR #143, opened 2021-03-01
Move nc vaccine scraper to tableau soure

sglyon:


sglyon:
@mikelehen and @ghop02 would you please take a look at how we mapped reported stats into our schema?

The dashboard is here: https://covid19.ncdhhs.gov/dashboard/vaccinations

The output of running the scraper at about noon eastern time was: 

```
In [61]: df.groupby("category")["value"].sum()
Out[61]: 
category
total_vaccine_completed     765630
total_vaccine_initiated    1301496
Name: value, dtype: int64
```

The troubling thing is that summing up their own numbers doesn't match either the CDC or what is reported in the summary tables for all of NC at the top of the dashboard. There is a disclaimer that data might be up to 72 hours late to arrive, which would then make sense that county data here is lower than the state totals (assuming it takes longer to figure out county stats than the agg)

mikelehen (approved):

cc7768 (approved):
lgtm

PR #146, opened 2021-03-03
ENH: Update variables we read from New Mexico API

cc7768:


sglyon (approved):

cc7768:
CDC comparison for first dose:

* state dashboard sum counties: 378,846
* cdc dashboard first doses: 488,139

This is probably a slightly larger difference than normal because I think the reservations are covered by the federal program rather than the state.

Worked through all of the numbers again and they seem okay.

PR #147, opened 2021-03-04
ENH add scraper for TX vaccines by demographic

sglyon:


cc7768 (approved):
The numbers seem to line up with what is reported on the dashboard.

Additionally, the county level aggregation for initiated comes relatively close to what is reported on the cdc page (3,705,059 vs 3,955,205)

PR #148, opened 2021-03-05
add scrapers for WI county and demographic vaccines

smcclure17:
dashboard source: https://www.dhs.wisconsin.gov/covid-19/vaccine-data.htm#summary

adds scrapers to track:
- cumulative `total_vaccine_initiated` and `total_vaccine_completed` at the county level.
- cumulative `total_vaccine_initiated` by race, age, sex, and ethnicity at the state level.

**_(see below)_** ~~This currently does not track "unknown" values. They are stored in separate tables than the rest of the data, which means (with the method I've used) a new scraper would need to be made for each. I can do this if it is important/wanted.~~ The proportion of "unknown" values for each section can be seen on the dashboard--at the moment they range from 0.5-8%. 

**_Added tracking for unknown and other categories for `race`, `sex`, and `ethnicity`_**. `Age` and the county-level data don't report missing/unknown entries, but they appear to exist.  The sum of all the `total_vaccine_initiated` entries for the `race`, `sex`, and `ethnicity` trackers all total `1088339`, which matches the corresponding value on the dashboard. For `age` and the county-level data these totals are lower (1088269 for `age` and 1081070 for the county-level). I don't know if this is normal/if there's an explanation for this, so I wanted to flag it. 

It seems to be hinted that there might be county level demographic data somewhere, (the demographical tables were labelled such as `... vax/unvax county`. But, the only data in these tables were state level, and there is no county level demographic data on their dashboard.


sglyon:
@mikelehen and @ghop02 what do you think about @smcclure17 question about handling "Unknown" vaccine recipients?

smcclure17:
Giving another look, the unknown values actually shouldn't be hard to add to the existing demographic scrapers. I could modify the `fetch()` function to return both the tables (the demographic table and its corresponding "unknown"table), then combine them. I believe they are structured the same way, so that concatenation should be easy (although I can't double check this until this evening).

mikelehen:
As a flag, the units here are different, right?
image

The race %s are the "% of that racial population that have been vaccinated" but unknown is the "% of vaccination records that didn't report race."

So:
1. If we did store `unknown`, the schema should reflect that these are two different kinds of measurements. What would that look like?
2. We probably don't have plans to surface the "unknown" data on the site.  It may be good to keep an eye on informally, just to get a sense of how good the coverage for state's demographic data is (and if it's unusually low, maybe we shouldn't show it), but I don't think we need to surface it, and so we can probably get by without collecting it.

mikelehen:
Ah, cool. If you're collecting cumulatives and properly annotating them, then it'd be great to leave them in.

sglyon:
Thanks team!

PR #153, opened 2021-03-09
Change CDC vaccine scraper variables

ghop02:
Ran this locally and looks like it works, however i don't think we should merge until we prepare for what we want to do for some of the states that have cumulatives going down. 

[trello ticket](https://trello.com/c/O70cNoYV/1069-vaccination-data-is-going-to-go-down-for-some-states-because-of-cdc-reporting-change) here


sglyon:
Thanks @ghop02 we'll wait to merge this one

PR #155, opened 2021-03-10
change GA county arcgis resource

tlyon3:
Original fix in PR #151 changed the ArcGIS resource for Georgia county. The ArcGIS server had updated the names and locations of the specific features, and the one that was originally used wasn't returning any data. 

This PR now changes back to the original resource as it is now returning data.

### DIFF
--- a/can_tools/scrapers/official/GA/ga_vaccines.py
+++ b/can_tools/scrapers/official/GA/ga_vaccines.py
@@ -25,7 +25,7 @@ class GeorgiaCountyVaccine(ArcGIS):
 
     def fetch(self):
         return self.get_all_jsons(
-            "Georgia_DPH_COVID19_Vaccination_public_v2_VIEW", 7, 6
+            "Georgia_DPH_COVID19_Vaccination_public_v2_VIEW", 6, 6
         )
 
     def normalize(self, data):


sglyon:
@tlyon3 this doesn't seem quite right:

```
d = GeorgiaCountyVaccine()

df = d.fetch_normalize()

df.query("location_name == 'Franklin'")[["dt", "category", "value"]]
Out[22]: 
            dt                          category  value
6   2021-03-10           total_vaccine_initiated   3136
165 2021-03-10           total_vaccine_completed   3828
324 2021-03-10  total_vaccine_doses_administered   6964
```

Notice that for this county vaccine_completed > vaccine_initated

This can't be true because our definitions are:

- vaccine_initiated: people with at least one dose
-  vaccine_completed: people with full series (2 for moderna/pfeizer or 1 for JJ)


Can you poke around the ArcGIS resources to see if we can get definitions for their columns so we can get these lined up properly?

tlyon3:
image
Looks like this is a data issue. I've looked at all the other counties and Franklin County is the only one that has this issue

sglyon:
Hey @mikelehen what do you think here? All the other counties have dose1 >= dose 2. It is just Franklin county. 

I tried goggling around for other resources directly from the county, but didn't find any. It looks like other sites/places are reporting what is in the picture above too

PR #156, opened 2021-03-11
Add scraper backfill

ghop02:
This adds a helper script to re-run normalize/validate/put for a specific scraper.  Lets you filter times to rerun.

Also, I modified the existing CDC vaccine scraper to use the Dose2 Recip number, which will let us backfill - then we can go back in with the series completed data


sglyon:
Nice, looks good! 

Did you run this already on the CDCStateVaccine scraper?

ghop02:
> Nice, looks good!
> 
> Did you run this already on the CDCStateVaccine scraper?

I downloaded all of the raw data from google cloud and ran it against my local instance.  Was thinking i'd run it on prod after it's on main

sglyon:
LGTM

PR #163, opened 2021-03-15
Adding a helper function to see latest vaccine data

ghop02:
Lets you run a specific scraper and see the vaccine data output:

image


sglyon (approved):
:tada:

PR #165, opened 2021-03-15
Add CT County Vaccine Scraper

ghop02:


can_tools/scrapers/official/CT/ct_vaccine.py
```diff
@@ -0,0 +1,50 @@
...
+                measurement="cumulative",
+                unit="people",
+            ),
+        }
+        out = data.melt(
+            id_vars=["dt", "location_name"], value_vars=crename.keys()
+        ).dropna()
+        out["value"] = pd.to_numeric(out.loc[:, "value"])
+        out["vintage"] = self._retrieve_vintage()
+        out = self.extract_CMU(out, crename)
```
sglyon, line 49: Need to drop `variable` column.

I've got it open, so I'll do that here

sglyon (approved):

sglyon:
@ghop02 looks great, thank you.

Two minor changes:

1. I dropped the `variable`  column. This doesn't actually break anything as our `put` methods just ignore it, but I did this to be consistent with the other scrapers
2. We had one missing `__init__.py` import in `can_tools/scrapers/__init__.py` that I added for you

PR #167, opened 2021-03-16
FEAT: Louisiana county vaccine scraper

tlyon3:


sglyon:
Good work, thanks!

Two small things:

1. let's drop the `variable` column before returning the dataframe. I usually do this using `result.drop(["variable"], axis="columns")`
2. Please make sure to start a new branch from current main when working on a different project. There aren't any merge conflicts, so we are ok this time, but we've been unlucky in the past.

PR #170, opened 2021-03-17
Adding MS County vaccine scraper and helper functions for easier data reshaping

ghop02:
Added the MS county vaccine data + a couple helper functions that may make it easier to get data frames in the right shape

### DIFF
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -1,10 +1,5 @@
 from can_tools.scrapers.base import CMU, DatasetBase
 
-# Is there any reason to pipe the imports through all the different layers? seems
-# like it's just as easy to add the one import directly to the scraper here rather than
-# having to include in a couple different __init__.py's.
-from can_tools.scrapers.official.MS.ms_vaccine import MSCountyVaccine
-
 from can_tools.scrapers.official import (  # IllinoisDemographics,; IllinoisHistorical,; Massachusetts,
     ArizonaData,
     ArizonaMaricopaVaccine,
@@ -89,6 +84,8 @@ from can_tools.scrapers.official import (  # IllinoisDemographics,; IllinoisHist
     WYCountyVaccinations,
     WYStateVaccinations,
 )
+
+from can_tools.scrapers.official.MS.ms_vaccine import MSCountyVaccine
 
 from can_tools.scrapers.ctp import (
     CovidTrackingProject,
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -1,4 +1,10 @@
 from can_tools.scrapers.base import CMU, DatasetBase
+
+# Is there any reason to pipe the imports through all the different layers? seems
+# like it's just as easy to add the one import directly to the scraper here rather than
+# having to include in a couple different __init__.py's.
+from can_tools.scrapers.official.MS.ms_vaccine import MSCountyVaccine
+
 from can_tools.scrapers.official import (  # IllinoisDemographics,; IllinoisHistorical,; Massachusetts,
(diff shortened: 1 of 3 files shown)


can_tools/scrapers/official/base.py
```diff
@@ -116,6 +116,116 @@ def _put_exec(self, engine: Engine, df: pd.DataFrame) -> None:
...
+            rename_columns[location_name_column] = "location_name"
+        if location_column:
+            rename_columns[location_column] = "location"
+        if date_column:
+            rename_columns[date_column] = "dt"
+
+        data = data.rename(columns=rename_columns)
+
+        if date_column:
+            data.loc[:, "dt"] = data["dt"].apply(lambda x: pd.to_datetime(x))
```
sglyon, line 212: Minor, but I think that having the rhs be `pd.to_datetime(data["dt"])` is same result, but potentially faster as it lets pandas vectorize

can_tools/scrapers/official/base.py
```diff
@@ -116,6 +116,116 @@ def _put_exec(self, engine: Engine, df: pd.DataFrame) -> None:
...
+            .pipe(self.extract_CMU, cmu=variable_map)
+            .drop(["variable"], axis=1)
+        )
+
+        if "vintage" not in data.columns:
+            data["vintage"] = self._retrieve_vintage()
+
+        return data
+
+    def _rename_or_add_date_and_location(
```
sglyon, line 166: I really like this, but am thinking maybe we split implementation into two functions?

Handling locations in a func and handling dates in a func seems like good modular design to me


If we wanted to reduce number of times we call both functions we could have a simple wrapper method that does both, similar to how `fetch_normalize` does `self.normalize(self.fetch())`

Another thought is that we could have these function arguments be class attributes instead and then we could write a general `normalize` method that could be applied to most scrapers?

can_tools/scrapers/official/MS/ms_vaccine.py
```diff
@@ -0,0 +1,94 @@
...
+import pandas as pd
+import camelot
+import requests
+import us
+
+from can_tools.scrapers import CMU
+from can_tools.scrapers.official.base import StateDashboard
+
+
+def normalize_table(
```
sglyon, line 13: Looks like this was moved into a helper method on DatasetBase, is that right?

can_tools/scrapers/__init__.py
```diff
@@ -1,4 +1,10 @@
 from can_tools.scrapers.base import CMU, DatasetBase
+
+# Is there any reason to pipe the imports through all the different layers? seems
+# like it's just as easy to add the one import directly to the scraper here rather than
+# having to include in a couple different __init__.py's.
+from can_tools.scrapers.official.MS.ms_vaccine import MSCountyVaccine
```
sglyon, line 6: Good question -- it is kinda annoying. I think we could probably just include it here as you outlined

sglyon (changes requested):
Overall I love what you are working on here 

Your work got me thinking about a couple other things we might to do extend your ideas and DRY even more.

can_tools/scrapers/official/base.py
ghop02: In general I agree though I'm not sure it's necessary in this specific case - I'm a proponent of "functions should do one thing well" and "renaming/adding date and time columns" feels like one thing but maybe I'm overstepping there.  If it was a bit more logic I'd probably be more in favor of splitting up.

I think a more general normalize is promising - but I'm still a bit hesitant to go there without using these helper functions more and seeing what would be useful.  For example I think the tableau class `normalize` is cool - but it does look like it's pushed some more of the work into `fetch` and I'm not sure that's the right place for it theoretically.

PR #176, opened 2021-03-17
FEAT: South Carolina PDF Vaccine Data

tlyon3:
There is a [tableau dashboard](https://scdhec.gov/covid19/covid-19-vaccination-dashboard), however I can't seem to find all the data. The PDF's are much more in depth with the available data.

The problem with the PDF's is that they aren't formatted consistently. I've made some headway with requesting all the data from the PDF's with a given date, but formatting it is proving to be a bit tricky.

Currently, the `fetch` method works, and will download the associated vaccine data for all three vaccines (Moderna, Pfizer, and Jenssen).


tlyon3:
The data for today probably hasn't been uploaded yet. By default, if the data isn't uploaded, I return an empty dataframe.

sglyon:
I see -- My preference would be to grab the most recent data. 

Could we do that?

sglyon (changes requested):
Change default behavior to grab most recent data

can_tools/scrapers/__init__.py
```diff
@@ -63,6 +63,8 @@
     PennsylvaniaVaccineEthnicity,
     PennsylvaniaVaccineRace,
     PennsylvaniaVaccineSex,
+    SCVaccineCounty,
+    SouthCarolinaTableauVaccineCounty,
```
sglyon, line 67: @tlyon3 is this old code now that you are using the pdf?

sglyon (changes requested):

can_tools/scrapers/__init__.py
tlyon3: changed

sglyon:
@tlyon3 this doesn't seem to be catching everything :(

Here's what I get from the scraper:

```
In [2]: d = SCVaccineCounty()

In [3]: data = d.fetch()

In [4]: df = d.normalize(data)
Normalizing pfizer data
Normalizing moderna data
Normalizing janssen data

In [5]: df.groupby("category")["value"].sum()
Out[5]: 
category
janssen_vaccine_completed     35554.0
moderna_vaccine_completed    118015.0
moderna_vaccine_initiated    173274.0
pfizer_vaccine_completed       5845.0
pfizer_vaccine_initiated      23454.0
Name: value, dtype: float64
```

And here are the summary stats in the Pfizer pdf:

Screenshot_20210331_115035

Looks like we aren't even close :(

tlyon3:
This is the problem that I talked about in [the discussion](https://github.com/covid-projections/can-scrapers/discussions/194). The missing data is probably coming from the pages that I couldn't figure out how to parse.

sglyon:
Ahh ok. Well for now I think we are stuck unless you can figure out how to parse those.

tlyon3:
PDF's are still not matching up. Luckily they are now displaying their data on a [Tableau dashboard](https://scdhec.gov/covid19/covid-19-vaccination-dashboard). This is great, because it ensures their are no parsing errors. However it doesn't seem to have all the data we want. It currently has:
* total doses by county
* at least one dose by county
* fully vaccinated _by state_

It is missing fully vaccinated by county. And I can't find a way to extrapolate the `fully vaccinated by county` with the current available data.

I've also found [their ArcGIS server](https://services2.arcgis.com/XZg2efAbaieYAXmu/ArcGIS/rest/services), but the data is the same data in the dashboard.

sglyon:
What's your recommendation ?

tlyon3:
I'm not sure. the PDF's have the most amount of data, but they are inconsistently formatted and difficult to parse via `camelot`. 

The dashboard doesn't have the data we want, and neither does the ArcGIS server.

sglyon:
What data is missing from Tableu and ArcGIS?

Do either  (or both together) of them provide people with at least one dose and people fully vaccinated by county?

tlyon3:
ArcGIS and  Tableau have the same data. They are both missing `total_vaccine_completed` _by county_. They have it by state, but not county

PR #177, opened 2021-03-18
ENH: Validators

cc7768:
Currently only adds validators used to compare values from a current scrape (not against the history)


sglyon:
@ghop02  here is the first stab at validator methods

Here's an overview:

- New `can_tools.validators` subpackage
    - There are three main files there 
        1. `utils.py` containing some handy utilities
        2. `cross_section.py` has validation helper functions that are applicable to data without a varyting time dimension
        3. `timeseries.py` has validatoin helper functions that require a varying time dimension  
- These helper functions are used in the new `validate` method on the `DatasetBase` class
- When I ran the validators I noticed that we had a number of issues
    - Some scrapers were broken and fixed(MT, TX, PA etc)
    - Some scrapers are broken and now have trello cards
    - Some scrapers failed validation and require decisions to be made. The summary of our findings about "invalid" data is in the discussion over here https://github.com/covid-projections/can-scrapers/discussions/204

can_tools/scrapers/base.py
```diff
@@ -402,6 +440,26 @@ def validate(self, df, df_hist):
         validated : bool
             Whether we have validated the data
         """
+        if is_time_series(df):
+            self._validate_time_series(df)
+
+        cats = [
```
ghop02, line 446: is the code here duplicating `self._validate_time_series`?

can_tools/scrapers/base.py
```diff
@@ -402,6 +440,26 @@ def validate(self, df, df_hist):
...
+            if cat_small in categories and cat_large in categories:
+                ok = cat1_ge_cat2(
+                    cumulatives,
+                    cat_large,
+                    cat_small,
+                    drop_levels=["unit", "category"],
+                )
+                if not ok:
+                    raise ValidateDataFailedError(f"{cat_small} >= {cat_large}")
         return True
```
ghop02, line 463: With validate functions I think we either want to always return a bool (i.e. true if no exception raised, or false if one is raised) or make it more of a `check` method and not return anything but raise an exception if there's a failed check.  Alternatively, I've seen examples of returning a list of errors encountered.  

I think one advantage of exceptions here is they'll be captured by sentry - but don't feel super strongly here

ghop02 (commented):
Took a quick first pass with some overall comments.  Looking good overall!

Couple additional notes:
 * would love to see some unit tests for the validators
 * Do we want to be able to override specific validators? not sure if we think it'll be an issue or not

can_tools/scrapers/base.py
```diff
@@ -383,6 +389,38 @@ def _normalize(self):
...
 
+    def _validate_time_series(self, df):
+        """Do checks only applicable to time series data"""
+        if (df["measurement"] == "cumulative").sum() > 0:
+            ok, problem = values_increasing_over_time(df)
+            if not ok:
+                import pprint
+
+                print("\n")
+                pprint.pprint(problem)
```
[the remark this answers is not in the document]
sglyon: Ahh oops, this was actually for my own testing. will delete these

can_tools/scrapers/base.py
```diff
@@ -383,6 +389,38 @@ def _normalize(self):
...
+                raise ValidateDataFailedError(
+                    f"{problem} has decrease in cumulative variable"
+                )
+
+    def _validate_order_of_variables(self, df):
+        cats = [
```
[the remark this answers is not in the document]
sglyon: I like this idea

can_tools/scrapers/base.py
sglyon: I value consistency. How would we return "false if one [exception] is raised"? Once we get an exception we give up control flow and can't provide a return value.

sglyon: I'm thinking maybe we should just not return anything?

can_tools/scrapers/base.py
sglyon: I refactored into `_validate_order_of_variables`, but forgot to delete here. Will do that now

sglyon:
Thanks for the quick review, I made some more changes.

I decided to go for having the validate method not return anything, but raise an exception.

I also made the helper _validate_XXX methods return a list of exceptions instead of throwing to make sure all issues are checked.

the _validate_XXX methods return instances of custom Exception subclasses that contain the details on what failed so we can print it out in a useful way

can_tools/scrapers/base.py
```diff
@@ -76,6 +114,15 @@ class DatasetBase(ABC):
     base_path: Path
     source: str
 
+    # this list of tuples specifies categories for which the first
+    # category listed should always be less than or equal to the second
+    # category listed, when the measurement type is cumulative
+    ordered_validation_categories = [
+        ("total_vaccine_initiated", "total_vaccine_doses_administered"),
+        ("total_vaccine_completed", "total_vaccine_doses_administered"),
+        ("total_vaccine_completed", "total_vaccine_initiated"),
```
mikelehen, line 123: Did we consider:

```
("total_vaccine_doses_administered", "total_vaccine_distributed")
```

?  Though I recall there being a bit of weirdness where we were sometimes getting 5 doses out of 4-dose vials or whatever...  so maybe it'll be noisy.  Happy to leave it out for now.

can_tools/scrapers/official/AZ/az_vaccine.py
```diff
@@ -184,7 +184,3 @@ def normalize(self, data) -> pd.DataFrame:
         out["dt"] = self._retrieve_dt("US/Arizona")
 
         return out.query("location_name not in @non_counties")
-
-    def validate(self, df, df_hist) -> bool:
-        "Quality is free, but only to those who are willing to pay heavily for it. - DeMarco and Lister"
```
mikelehen, line 189: 😂

can_tools/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    """Loads most recent cdc data either from scraper output, sql, or parquet file
+
+    Args:
+        source: Where to get data. One of "scraper_output", "sql", "parquet"
+        engine: sqlalchemy engine for sql queries. Only necessary when source = "sql"
+
+    Returns:
+        pd.DataFrame: DataFrame with CDC Covid Tracker data
+    """
+    pass
```
mikelehen, line 44: What's up with this function?  WIP?

can_tools/scrapers/base.py
```diff
@@ -41,6 +47,38 @@ class ValidateDataFailedError(Exception):
     """Error raised when data vailidation fails."""
 
 
+class ValidateRelativeOrderOfCategoriesError(Exception):
+    def __init__(self, cat_small, cat_large, problems):
```
mikelehen, line 51: FWIW- I have trouble reading this as "category" since "cat" has a very well-defined meaning in my head already... and while it's fun to have large and small cats in the code, it may not lend to readability. 😁 

Furthermore, why are these "categories" and not "columns" or "variables" or something?  Are things like "total_vaccine_initiated" considered a category in the codebase?

can_tools/validators/timeseries.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class CumulativeValueDecreases(Exception):
+    def __init__(self, info):
+        self.info = info
+
+
+def values_increasing_over_time(df: pd.DataFrame) -> Dict[str,Any]:
+    """
+    Check if all cumulative variables are always
```
mikelehen, line 14: comment fragment

can_tools/validators/timeseries.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Args:
+        df: normalized Dataframe with varying `dt`
+
+    Returns:
+        bad: A dictionary mapping from (category, age, sex, race, ethnicity, sex)
```
mikelehen, line 20: `sex` is in here twice...

mikelehen (approved):
Left some drive-by comments which you're welcome to ignore as desired, but this looks like a great start.  Thanks for getting this going!

can_tools/utils.py
sglyon: Yeah sorry, that's WIP -- it is meant to be the starting point/API spec for working towards this https://trello.com/c/Htnq78Bw/987-test-we-should-fail-if-a-county-level-scraper-aggregated-to-the-state-does-not-agree-with-the-cdc-within-5-point

can_tools/scrapers/base.py
sglyon: Good questions.

Happy to spell out category.

And yes, we call a "variable" a triplet of (category, measurement, unit). So something like (total_vaccine_initiated, cumulative, people) is a "variable"

It isn't column because data all long form, so the data are all in the  `value` column

can_tools/validators/timeseries.py
sglyon: no comment

can_tools/scrapers/base.py
sglyon: Good thought. I didn't think about that one because we don't have a whole lot of `total_vaccine_distributed` data

Probably (I hope!) wouldn't hurt to include because any scrapers that don't have both variables would just move past that check

can_tools/scrapers/base.py
```diff
@@ -383,37 +431,73 @@ def _normalize(self):
         df = self.normalize(data)
         return self._store_clean(df)
 
+    def _validate_time_series(self, df) -> [List[Exception]]:
+        """Do checks only applicable to time series data"""
+        issues = []
+        if (df["measurement"] == "cumulative").sum() > 0:
```
ghop02, line 437: would `(df["measurement"] == "cumulative").any()` work?

can_tools/scrapers/base.py
```diff
@@ -383,37 +431,73 @@ def _normalize(self):
...
-            Whether we have validated the data
         """
-        return True
+        issues = []
+        if utils.is_time_series(df):
+            issues.extend(self._validate_time_series(df))
+
+        issues.extend(self._validate_order_of_variables(df))
+
+        if len(issues) > 0:
```
ghop02, line 493: nit: would prefer `if issues:` [gsg explanation](https://google.github.io/styleguide/pyguide.html#2144-decision)

ghop02 (approved):
This seems good to me! I say we ship and iterate if there are parts that we don't love

can_tools/scrapers/base.py
sglyon: Yep that would probably work

can_tools/scrapers/base.py
sglyon: I think I'm going to go against gsg on this one.


The first phrase

> Use the “implicit” false if possible,

grates at me relative to the zen of python

> Explicit is better than implicit.

PR #181, opened 2021-03-20
Uses new california spreadsheet of county level vaccination data

ghop02:


sglyon (approved):
This is awesome! 

Thank yo

PR #182, opened 2021-03-20
Cdc scraper single state

ghop02:
Modifies the CDC scraper to be able to run for a single state rather than having to run on all of them.

This creates a new prefect flow which kicks off a bunch of single state CDC scraper runs. Hopefully it makes our database a bit happier

### DIFF
--- a/services/prefect/flows/generated_flows.py
+++ b/services/prefect/flows/generated_flows.py
@@ -5,19 +5,21 @@ import pandas as pd
 import sqlalchemy as sa
 from can_tools import ALL_SCRAPERS
 from can_tools.scrapers.base import DatasetBase
-
+from can_tools.scrapers.base import ALL_STATES_PLUS_DC
+from can_tools.scrapers import CDCCovidDataTracker
 import prefect
 from prefect import Flow, task
 from prefect.schedules import CronSchedule
 from prefect.tasks.secrets import EnvVarSecret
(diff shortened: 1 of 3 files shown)


services/prefect/flows/generated_flows.py
```diff
@@ -93,8 +95,58 @@ def create_flow_for_scraper(ix: int, d: Type[DatasetBase]):
...
+
+    return flow
+
+
+def create_cdc_all_states_flow():
+    """Creates a flow that runs the CDC data update on all states."""
+    sched = CronSchedule("17 */4 * * *")
+
+    flow = Flow("CDCAllStatesDataUpdate", sched)
+    for state in ALL_STATES_PLUS_DC:
```
sglyon, line 124: Chris, I think this would start all states at the same time. 

Do you know if there is a way to do it batches, say start 5/wait, start next 5/wait, etc.?

sglyon (changes requested):
This is awesome!

I I had one question...

services/prefect/flows/generated_flows.py
ghop02: When I was running it locally it was running one after the other - so I don't think it should kick all off at once (I think the wait=True does that)

sglyon: ooh fancy -- I like that

sglyon (approved):

PR #183, opened 2021-03-21
fixes for Texas vaccine demographic scrapers

smcclure17:
Small changes to column/key names.

`People Vaccinated  ` appears to be people vaccinated with at least one shot, but I could not find a proper/formal definition.


sglyon:
Awesome, thakn you

PR #184, opened 2021-03-21
DC Demographic Vaccine Scraper (Race)

smcclure17:
source: https://coronavirus.dc.gov/data/vaccination

They surface demographic data for sex/age/ethnicity, but I couldn't find it in the returned JSON. I'll look more tomorrow/Monday (3/21/21).

For now, I wanted to open this just to have something up

### DIFF
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -19,6 +19,7 @@ from can_tools.scrapers.official import (  # IllinoisDemographics,; IllinoisHist
     DCCases,
     DCDeaths,
     DCGeneral,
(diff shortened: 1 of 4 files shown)


sglyon:
Thanks @smcclure17 

This was a scraper for sex, so I made that change. Looking forward to more!

smcclure17:
hmm, the `Demographics`  table was (i'm pretty sure) returning data for race before. I also think that the dashboard's default view was data by race, instead of by sex as it is now, which might have something to do with it. 

I can keep an eye on it, and keep looking for the other data

PR #186, opened 2021-03-23
correct vaccine variables

sglyon:
- fix MD vaccine scraper
- clean OH vaccine
- cleanup nc_vaccine variable names
- fix nj vaccine
- clean ny vaccine scraper variables
- clean column names for oregon county vaccine scraper
- clean some column names for nm vaccine
- use common variables for ct_vaccine
- use common variable names for az_vaccine
- fix mn vaccine scraper
- remove incorrect variable from MT scraper
- Fix definition of IN vaccine variables
- use common cmu defs for WI vaccine
- use common cmus for wa vaccine"
- account for J&J in MI scraper
- use shared cmu for mo scraper
- use shared variable in MO


sglyon:
@ghop02 this contains the work to make sure scrapers properly account for J&J data

About 1/2 the commit make a change to a scraper b/c we were doing things wrong.

The other 1/2 are cosmetic -- most of them to use the shared CMU definitions in variables.py

can_tools/scrapers/official/AZ/az_vaccine.py
```diff
@@ -1,3 +1,8 @@
+from can_tools.scrapers.variables import (
+    FULLY_VACCINATED_ALL,
+    INITIATING_VACCINATIONS_ALL,
+    TOTAL_DOSES_ADMINISTERED_ALL,
+)
```
ghop02, line 5: Nit, but I'd advocate for importing like `from can_tools.scrapers import variables` and then referring to them by `variables.FULLY_VACCINATED_ALL` etc. 

https://google.github.io/styleguide/pyguide.html#22-imports has a good explanation - I like that it gives more context to where some constants are coming from - and you don't have to add a new import for every new variable you use

ghop02 (changes requested):
Overall looks good! It's small, but I think change the import structure as mentioned in my first comment would make it a bit cleaner overall

can_tools/scrapers/official/AZ/az_vaccine.py
sglyon: Makes sense

can_tools/scrapers/official/MD/md_vaccine.py
```diff
@@ -18,62 +23,26 @@ class MarylandCountyVaccines(ArcGIS, DatasetBase):
...
-            "measurement",
-            "unit",
-            "age",
-            "race",
-            "ethnicity",
-            "sex",
-            "value",
-        ]
-        return out.loc[:, cols_to_keep]
+        return df.pipe(
```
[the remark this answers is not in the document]
sglyon: I agree on this one - I'll split it

can_tools/scrapers/official/MD/md_vaccine.py
```diff
@@ -18,62 +16,32 @@ class MarylandCountyVaccines(ArcGIS, DatasetBase):
...
-            )
-            .dropna()
+        # singledose was added only after J&J vaccines started.
+        # If it appears it will only be reported in the fully vacinated
+        # column. We also wanted to include it in `firstdose` b/c we
+        # report "people with at least one dose"
+        if "singledose" in list(df):
+            df["firstdose"] += df["singledose"]
+
+        out = pipe(
```
ghop02, line 39: should this just be `out = self._rename_or_add_date_and_location(...)`?

sglyon: they are equivalent. If we are doing it this way (just one method, not in a chain), then your proposal makes sense to me.

I'll make a change

ghop02 (approved):

PR #189, opened 2021-03-24
add race, ethinicity, age scrapers for DC vaccines

smcclure17:
concept for scraping data from subsheets using[ `TableauScraper`](https://github.com/bertrandmartel/tableau-scraping/blob/master/README.md) library. 

data source: https://coronavirus.dc.gov/data/vaccination

---
note: 

When trying to access the `Gender` column/subsheet an error was thrown,  despite it appearing to exist (See below). So, I left the `sex` scraper as it is currently, but I can look into what the problem is if wanted.

gender
error


sglyon:
Awesome thanks for this @smcclure17 

I know that others including @christopherturner are interested in this functionality. 

Should we hold off reviewing until after you sync with them?

smcclure17:
Yeah, I think that's smart--we are planning to sync tomorrow

can_tools/scrapers/official/DC/dc_vaccines.py
```diff
@@ -13,7 +14,7 @@ class DCVaccineSex(TableauDashboard):
     location_type = "state"
     baseurl = "https://dataviz1.dc.gov/t/OCTO"
     viewPath = "Vaccine_Public/Demographics"
-
+    demographic_cmu = "sex"
     data_tableau_table = "Demographics "
```
tlyon3, line 19: adding `demographic_col_name = "Gender"` here doesn't seem to break the scraper. Is that necessary? I noticed that `demographic_col_name` exists in the other demographic scrapers

smcclure17: Yes, it shouldn't hurt to add. Since the sex/gender data is returned by default (without any further filtering), I didn't use the `TableauScraper` library to extract the data, which is what the `demographic_col_name` is used for. 

I'm not sure if we're going to use this library at all, @christopherturner has some code that does filtering/selection manually (in our `TableauDashboard` class), so I think we might/will/should restructure these scrapers to use that

tlyon3:
@smcclure17 could you resolve merge conflicts with `main` and re-request a review?

smcclure17:
@tlyon3 yes, will do. There are some hefty merge conflicts that I will look at tomorrow (4/6/2021)--I ran out of time today

smcclure17:
the data now return `unknown` data for race and ethnicity automatically, rendering the `_get_unknown()` function useless. But, I've left its definition in, in case it is needed in the future. 

`unknown` data for Age and Sex are not reported in these queries. But, there is little unknown data for these categories (Sex has < 600 unknown entries, and Age has < 50). If wanted, I could look into finding ways to collect this data, but at this point I have not.

tlyon3 (changes requested):
Getting an error in the `_get_date` method. There is no `Update` key in the tableau view. Could you look into that?

smcclure17:
hmm I don't seem to be getting the same issue. But, the key the `_get_date()` method should be using is `Admin Update`--when this scraper was first made that key was `Update` but the name of the table was changed. When changing the key to `Update` I do get that same error. 

This branch looks like it's using the `Admin Update` key (line 31 in `dc_vaccines`) and it works when I run it. If you could check that you have the right key (`Admin Update`) in the function, then if the error persists I can take another look.

tlyon3:
Ah ok. Yes it is working now. I think I must've not pulled to get the most recent. My bad

tlyon3 (approved):

sglyon (approved):

PR #190, opened 2021-03-24
Fix Missouri vaccine values

ghop02:
Looks like they changed their values


sglyon:
Hmm odd. I made this change yesterday to match their dashboard. 

Did they really switch names again that quickly?

ghop02:
> Hmm odd. I made this change yesterday to match their dashboard.
> 
> Did they really switch names again that quickly?

it seems they did - maybe the switch yesterday was unintentional :/

PR #191, opened 2021-03-25
FIX: Data doesn't represent J&J vaccine accurately

tlyon3:
Some of the data providers don't report the J&J vaccine following our definition. They only include the J&J vaccine counts in their second dose column. 

This pull request simply removes the `total_vaccine_initiated` column until either we can figure out how to separate the J&J dose counts from other vaccines, or the data provider changes the way they report.


sglyon:
Hey @tlyon3 thanks for doing this as asked.

I regrouped with the rest of the team and we decided to continue to provide the data as -is (even though we know it is wrong) and put a discliamer/note on the website saying we might be underestimating the number of people with at least one dose

We'll also keep an eye on these dashboards and hope they fix things up soo

PR #192, opened 2021-03-25
clean unused scrapers

sglyon:
- wip on cleaning scrapers
- fix column names for TX vaccine demographics scraper
- update service name for AL vaccine
- remove engine="xlrd" (deprecated) from MI vaccine
- clear out init files
- isort
- black


sglyon:
Hey @ghop02 this cleans house.

I deleted scrapers that we don't need anymore (only provide cases | deaths | tests | hospitalization with demographic data), reorganized the imports, fixed AL vaccine, FL vaccine, Mi vaccine and ran isort + black

scratch.py
```diff
@@ -0,0 +1,20 @@
+import asyncio
```
ghop02, line 1: Was this intentional?

ghop02 (approved):
This looks good to me!

scratch.py
sglyon: whoops, now

PR #197, opened 2021-03-29
FIX: Fix, NCVacccine, IowaCountyVaccine, MontataCountyVaccine, CTCountyVaccine, Pennsylvania

tlyon3:
NCVaccine:
* updated column names

IowaCountyVaccine:
* updated token

MontanaCountyVaccine:
* the names of the counties were incorrect. The data includes the fips numbers so I changed the scraper to use those instead

CTCountyVaccine:
* Removed non-county entry in data

Pennsylvania Scrapers:
* updated source


sglyon:
Awesome, thank you!

PR #201, opened 2021-03-30
add SD county vaccine scrapers

smcclure17:
scrapes `total_vaccine_initated` and `total_doses_administered` for South Dakota counties from https://doh.sd.gov/COVID/Dashboard.aspx

### DIFF
--- a/can_tools/scrapers/official/SD/sd_vaccines.py
+++ b/can_tools/scrapers/official/SD/sd_vaccines.py
@@ -1,0 +1,279 @@
+import json
+
+from typing import Any
+
+import pandas as pd
+import us
+
+from can_tools.scrapers import CMU, variables
+from can_tools.scrapers.util import flatten_dict
+from can_tools.scrapers.official.base import MicrosoftBIDashboard
+
+"""
+    Notes:
+    J&J shot included in "one shot" data, but NOT in two shot data
+"""
+
+
+class SDVaccineCounty(MicrosoftBIDashboard):
+
+    has_location = False
+    location_type = "county"
+    state_fips = int(us.states.lookup("South Dakota").fips)
+
+    source = "https://doh.sd.gov/COVID/Dashboard.aspx"
+    source_name = "South Dakota Department of Health"
+    powerbi_url = "https://wabi-us-gov-iowa-api.analysis.usgovcloudapi.net"
+
+    variables = {
+        "at_least_one_dose": variables.INITIATING_VACCINATIONS_ALL,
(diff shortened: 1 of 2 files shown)


can_tools/scrapers/official/SD/sd_vaccines.py
```diff
@@ -0,0 +1,279 @@
...
+            {
+                "Query": {
+                    "Commands": [
+                        {
+                            "SemanticQueryDataShapeCommand": {
+                                "Binding": {
+                                    "DataReduction": {
+                                        "DataVolume": 3,
+                                        "Primary": {"Window": {"Count": 100}},
+                                        "Secondary": {"Top": {"Count": 100}},
```
tlyon3, line 53: Do you know if this is limiting the number of rows returned?

smcclure17: that does look like it might. This query returns all of the data completely, it was a query I made using the `MicrosoftBIDashboard` `_construct_...()` functions that was not returning all the data. 

When I increased those values (to 1000) the response had all the same data, but also returned a warning: 
`[{'Code': 'SpecifiedReductionAlgorithmsExceedsMaxIntersections', 'Severity': 'Warning', 'Message': 'Specified reduction algorithms exceed the maximum allowed number of intersections (1000000). The counts will be ignored.'}]`

tlyon3: Hmm...ok. Do the other PowerBI scrapers include this "filter" as well?

smcclure17: I've found similar filters in some queries of other BI dashboards, but it is not universal. Below is an example of a query from the Minnesota dashboard. I haven't found a pattern as to why it's included in some instances and not others--but I can keep searching.

image

tlyon3: If you haven't found anything contradictory, and the totals are matching the dashboard, then I'll approve this

smcclure17: @tlyon3 the values match, so this should be good to go. (only scrapes people initiating vaccine and total doses administered)

tlyon3 (approved):

sglyon (approved):
Looks great, thanks.

Is there anyway to get that table breaking out the doses by shot and number in sequence?

Screenshot_20210406_104233

smcclure17:
This is the data that I was trying to collect in #199. The query posted there returns the data from that table for about 1/3 of the counties.

PR #202, opened 2021-03-30
FEAT: alaska scraper

tlyon3:
I've formatted the data but now I'm stuck. There are a few data questions I have and I'm not sure how to proceed.

Some of the rows have a `dose_number == 3`. I'm not sure how to interpret that. My initial though was those were J&J doses. But that doesn't seem to match up with the data. I've labeled the corresponding columns with `dose_unknown`.

Some rows had a `fips` value of `000` or `NULL`. I assume that `NULL` means unknown county. I drop those rows. I thought `fips == '000'` means statewide, but that doesn't match up.

Here is the dashboard for reference: https://alaska-coronavirus-vaccine-outreach-alaska-dhss.hub.arcgis.com/app/c74be37e02b44bb8b8b40515eabbab55


smcclure17 (approved):

can_tools/scrapers/official/AK/ak_vaccine.py
```diff
@@ -0,0 +1,262 @@
...
+        )
+        
+        out = (out.pipe(self.extract_CMU, cmu=self.variables)
+            .assign(
+                vintage=self._retrieve_vintage(),
+                dt=out['dt'].dt.date
+            )
+            .drop(["variable"], axis=1)
+        )
+        return out.rename(columns={'fips':'location'})
```
smcclure17, line 137:
```suggestion
        out = out.rename(columns={'fips':'location'})
        out = out.query("location != 2000").replace(2270, 2158)
        return out
```
The 2270 fips was renamed and now has a fips of 2158, so this updates that change. 

removed the `000` county so that the tests would pass, but not sure what the best course of action is.

smcclure17 (commented):
We are still unsure how to handle the `000` county and `dose_unknown` values, but otherwise the scraper is good to go. Currently we don't include the `dose_unknown` data and the `000` fips does not match a location, causing the tests to fail. I've removed it in one of the suggestions, but I don't know what the best course of action is

sglyon:
Hey @tlyon3 sorry for super long delay on this one.

I tried running `AlaskaCountyVaccine`  today and saw that it is broken :(

It complains that a column `primarkeyid` does not exist in `fetch`ed data.

Perhaps we could switch to using `objectid` column instead? Will you please take a look and see if we can get it running.

tlyon3:
Changed and is running

PR #203, opened 2021-03-30
fix/update DC demographic vaccine scraper

smcclure17:
temporarily updated scraper to match the data that the dashboard returns, (changed from Sex data to Race data)

Also, the previous _partially vaccinated_ data did not include people who had completed their series. This should fix this by summing the partially vaccinated people with fully vaccinated to match the `total_vaccine_initiated` definition. Sorry I didn't catch that earlier.

Added a check that should stop the scraper from running if the demographic changes (eg, if it switches back to data by sex)

### DIFF
--- a/can_tools/scrapers/official/DC/dc_vaccines.py
+++ b/can_tools/scrapers/official/DC/dc_vaccines.py
@@ -2,10 +2,11 @@ import pandas as pd
 import us
 
 from can_tools.scrapers.base import CMU
+from can_tools.scrapers import variables
 from can_tools.scrapers.official.base import TableauDashboard
 
 
-class DCVaccineSex(TableauDashboard):
+class DCVaccineRace(TableauDashboard):
     has_location = True
     source = "https://coronavirus.dc.gov/data/vaccination"
     source_name = "DC Health"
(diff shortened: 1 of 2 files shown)


sglyon:
@tlyon3  please take a look

tlyon3 (approved):

can_tools/scrapers/official/DC/dc_vaccines.py
```diff
@@ -17,49 +17,57 @@ class DCVaccineSex(TableauDashboard):
...
-        out = df.assign(sex=df["Cross-value"].str.lower()).drop(
-            columns={"Cross-value", "variable"}
+        # sum the partially and fully vaccinated entries to match definition
+        # to avoid pivoting to wide then back to long I selected each corresponding entry to sum
+        q = 'variable == "{v} VACCINATED" and demo_val == "{d}"'
+        for d in ["black", "white", "other"]:
+            initiated_value = int(
+                df.query(q.format(v="PARTIALLY", d=d))["value"]
+            ) + int(df.query(q.format(v="FULLY", d=d))["value"])
+            row = {"variable": "INITIATED", "demo_val": d, "value": initiated_value}
```
sglyon, line 58: Let's also add the "unknown" data here

I think the values for (age, race, ethnicity, sex) would then be (all,unknown,unknown,all)

sglyon (changes requested):

can_tools/scrapers/official/DC/dc_vaccines.py
smcclure17: new commit should add this. I had to do some calculations but it looks to be correct

sglyon (approved):

PR #214, opened 2021-04-02
ENH: Add CDC county vaccine data

cc7768:
@ghop02 @sglyon 

Popping in to add to your review load.

### DIFF
--- a/can_tools/scrapers/base.py
+++ b/can_tools/scrapers/base.py
@@ -37,8 +37,14 @@ class CMU:
         self.sex = sex
 
 
+class RequestError(Exception):
+    pass
+
+
 class ValidateDataFailedError(Exception):
     """Error raised when data vailidation fails."""
+
+    pass
 
 
 def _get_base_path() -> Path:
--- a/can_tools/scrapers/base.py
+++ b/can_tools/scrapers/base.py
@@ -38,6 +38,8 @@ class CMU:
 
 
 class RequestError(Exception):
+    """Error raised when a network request fails"""
+
     pass
 
 
--- a/can_tools/scrapers/official/federal/CDC/cdc_county_vaccines.py
+++ b/can_tools/scrapers/official/federal/CDC/cdc_county_vaccines.py
@@ -52,7 +52,7 @@ class CDCCountyVaccine(FederalDashboard):
         if not response.ok:
             msg = f"Failed to make request to {self.url}\n"
             msg += "Response from request was\n:"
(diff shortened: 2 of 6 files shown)


sglyon (approved):

can_tools/scrapers/base.py
```diff
@@ -37,9 +37,17 @@ def __init__(
         self.sex = sex
 
 
+class RequestError(Exception):
```
ghop02, line 40: suggestion/thought for the future: as long as there are more than a few I've always liked having an `errors` module that makes it easy to reference common errors across different contexts.

can_tools/scrapers/official/federal/CDC/cdc_county_vaccines.py
```diff
@@ -0,0 +1,64 @@
...
+            age="65_plus",
+        ),
+    }
+
+    def fetch(self):
+        response = requests.get(self.url)
+        if not response.ok:
+            msg = f"Failed to make request to {self.url}\n"
+            msg += "Response from request was\n:"
+            msg += textwrap.indent(response, "\t")
```
ghop02, line 55: Should this be `response.content`?

can_tools/scrapers/variables.py
```diff
@@ -2,31 +2,31 @@
 
 from can_tools.scrapers import CMU
 
-INITIATING_VACCINATIONS_ALL = CMU(
+INITIATING_VACCINATIONS_ALL: CMU = CMU(
```
ghop02, line 5: Is this necessary when the variable is directly assigned?

ghop02 (approved):
Couple small comments!

can_tools/scrapers/base.py
sglyon: Yeah that works for me. so far I've just been collecting them here

can_tools/scrapers/official/federal/CDC/cdc_county_vaccines.py
sglyon: yep, good catch!

can_tools/scrapers/variables.py
sglyon: For some reason my editor wasn't picking up that it was a CMU instance when I was using `variables.INITIATING_VACCINATIONS_ALL`

I agree it shouldn't be necessary, but I guess for my tooling it was?

ghop02:
@sglyon is this good to go?

PR #217, opened 2021-04-04
Disable USAFacts validation

ghop02:


mikelehen (approved):
LGTM, assuming it works. 😁

PR #218, opened 2021-04-04
Florida demographis data

mehanig:
This PR reads data from http://ww11.doh.state.fl.us/comm/_partners/covid19_report_archive/vaccine-county/vaccine_county_report_latest.pdf

And tries to create one dataframe with merged values, but I am not sure I understand the final schema should look like

Dataframe with only age data looks like this (dfs_demographics variable at line 48):
image

Can someone give me advice how this should be merged into final Dataframe.
I commened the code which supposed to do something similar, but it doesnt work and I want to have a feedback before I dig deeper into it


mehanig:
@sglyon saw your reply in another PR, thanks for explaining with long data, I will apply the comment to my PR too

sglyon:
Thank you for looking and responding here! 

Sorry I didn’t get a chance here. 

I have been extra tied up with external things this week. 

I’m working on some developer documentation that will hopefully make it easier to contribute. 

Thanks for your patience and willingness to help us.

sglyon:
The WIP docs for writing scrapers are here: https://github.com/covid-projections/can-scrapers/blob/docs/docs/scraping/writing_scrapers.rst

I'll attempt to get these "published" by EOD Friday

sglyon:
update -- scope of docs expanded a bit and it is now slated to be released by end of next week!

mehanig:
Hello, thank you for docs! 
They were helpfull. I pushed an update which creates dataframe in a format needed, I believe. 
Code is a bit repetetive, but let's focus on data format for now.

Since I am changing the existing file, I coudn't find a way to merge existing dataframe and demographic dataframe, is it better be a new file? Or there is an expected way to merge them?

Another question is that test is failing with `"NOT NULL constraint failed: covid_observations.location_id"`, What's needed to be done with location_id? I somehow missed this part, do I need to add location to variable files?

Thank you

sglyon:
Hey @mehanig thanks for the work here. 

> Since I am changing the existing file, I coudn't find a way to merge existing dataframe and demographic dataframe, is it better be a new file? Or there is an expected way to merge them? 

It's ok to have these scrapers live in the same file, but we won't need to worry about merging the data from the other scraper from the demographic data here. Let's keep them in two separate classes and let the database generate the combined dataset for us.

> Another question is that test is failing with "NOT NULL constraint failed: covid_observations.location_id", What's needed to be done with location_id? I somehow missed this part, do I need to add location to variable files?

You shouldn't have to add anything to the variable files. The general rule of thumb with the location Id is that your dataframe should have exactly one of the following:

1. A column named `location` containing county FIPS codes as integers AND you should set the class level property `has_location = True`
2. A column names `location_name` containing the name of a county as a string AND setting `has_location = False` at the root of the class

@tlyon3 will you please review  this PR on Wednesday morning and give any more guidance on the location_id issues 

(note I'm OOO for the next couple days, so @tlyon3  is helping pick up my slack!)

can_tools/scrapers/official/FL/fl_vaccine.py
```diff
@@ -13,18 +16,77 @@ class FloridaCountyVaccine(StateDashboard):
...
+                f.write(results.content)
+                pdf_pages_headers = textract.process(tmp_file)
+                county_names = re.findall(r"COVID-19: (?P<countyName>.*?) vaccine summary",
+                                          pdf_pages_headers.decode("utf-8"))
+                county_names = list(map(lambda x: x if not x.endswith(" County") else x[:-7], county_names))
+
+        county_demographics_data = camelot.read_pdf(self.fetch_url_for_counties, pages="1-end", flavor="stream", row_tol=10)
+        return {"vaccine_data": vaccine_data, "county_demographics_data": county_demographics_data, "county_names": county_names}
+
+    def normalize_group(self, df, demographic, dgroup, group_rename, cmu):
```
tlyon3, line 41: Looks like this method is unused? You normalize each demographic data in `if` blocks in the main `normalize` method. 

I'd recommend breaking them out into individual demographic normalize methods (eg `_normalize_sex`, `_normalize_age`, etc). Then call each one in the main `normalize ` method.

I believe that's what you were starting to do with this method.

tlyon3 (changes requested):
Looks good. Looks like got the `location_id` error fixed. Good job! Just one comment on code structure and this is good to go.

sglyon:
Hi @mehanig just checking in to see if you've had any luck fixing the SQL issues with the location columns?

Would it help to have a brief call? If so, please reach out on slack

sglyon:
Hey @smcclure17 I have gone through this PR and made changes so that it now works!

Would you please review when you get a chance?

Thanks!

smcclure17 (approved):
tests passed and the data looks good/correct!

I don't have too much experience with parsing from PDFs, so no real comments there, but I think everything looks great!

PR #219, opened 2021-04-04
WIP Add NC Demographics

kxdan:
This PR collects information from NC statewide on Race, Ethnicity, Gender and Age breakdowns from North Carolina.

I'm not entirely sure of data format we're after here, and would appreciate insight as to how to get it in the correct format. Here's an example of what the data looks like for age

image

For now, this PR only adds state-wide collection of information, as the other worksheets with county-by-county data require some special parsing logic that i've not yet looked into, once this is integrated, I will look further into this


sglyon:
Hi Thanks @kxdan!

The structure we are looking for is written in the readme [here](https://github.com/covid-projections/can-scrapers#writing-a-new-scraper)

To reiterate a little bit, the `fetch` method should do the network requests necessary to get data from remote resource, but shouldn't bother with parsing/cleaning/reshaping (if data already comes parsed like it seems to from the TableauScraper class used here that's k)

The `normalize` method should return data in [long form](https://www.theanalysisfactor.com/wide-and-long-data/). The columns are written in the readme. I'll repeat here: `(vintage, dt, location, category, measurement, unit, age, race, ethnicity, sex, value)` Note that all columns except `value` are "metadata" that describe what is being recorded.

For the case with the age data you presented above, here's what this might look like:

- vintage: recorded automatically for you
-  dt: this is the date of observed data -- derived from "Week of" column in your screenshot
-  category, measurement, unit: comes from one of the rows in [this csv file](https://github.com/covid-projections/can-scrapers/blob/main/can_tools/bootstrap_data/covid_variables.csv)
- age: this is the key distinguishing dimension for this dataset. It would need to come from the `Age group (copy)-alias` column in your screenshot, but be parsed and match one of the rows in the [covid_demographics.csv file](https://github.com/covid-projections/can-scrapers/blob/main/can_tools/bootstrap_data/covid_demographics.csv). If there does not exist a suitable row in that csv, you would add one. For example, if the age group reported on teh dashboard were something odd like 33-37 and we didn't have an entry for that in the csv, you'd need to add `33-37,all,all,all` to the csv file
- race,ethnicity,sex: these will all be equal to `"all"` as we are not differentiating on this dimension right now
- value: this is the actual count of the (category, measurement, unit) variable for the (age, race, ethnicity, sex) demographic group. From what I can tell it looks like this is contained in the far right column of your screenshot.

I hope this provided some help, if not, please reach back out!

sglyon:
I'm working on some developer docs I hope to publish by EOD friday

You can find the draft here: https://github.com/covid-projections/can-scrapers/blob/docs/docs/scraping/writing_scrapers.rst

kxdan:
So i've added some changes that should put the data in the format requested, my colleague Mike and I have the same issue in terms of an SQL issue when running locally (location_id NOT NULL), any clue?

sglyon:
This is looking great. Thank you for the work here!

@tlyon3  will you please review this PR on Wednesday morning and advise @kxdan  on the missing location id in SQL?

Great work!

can_tools/scrapers/official/NC/nc_vaccine.py
```diff
@@ -1,8 +1,10 @@
 import pandas as pd
+from pandas.core.frame import DataFrame
 import us
-
+import datetime as dt
 from can_tools.scrapers.base import CMU
 from can_tools.scrapers.official.base import TableauDashboard
+from tableauscraper import TableauScraper as TS
```
tlyon3, line 7: The `TableauDashboard` superclass does have methods for extracting data from a tableau dashboard. The other scrapers in the repo are using those tools. I believe we want to utilize that instead of the `tableauscraper` package. @sglyon, can you confirm?

can_tools/scrapers/official/NC/nc_vaccine.py
```diff
@@ -15,7 +17,7 @@ class NCVaccine(TableauDashboard):
     location_type = "county"
     baseurl = "https://public.tableau.com"
     viewPath = "NCDHHS_COVID-19_Dashboard_Vaccinations/Summary"
-
+    location_name = "NorthCarolina"
```
tlyon3, line 20: The name in the database is `North Carolina` This will cause the problem with the "missing `location_id`" error. 

However, I changed it locally and I'm still getting the error. Still investigating.

can_tools/scrapers/official/NC/nc_vaccine.py
```diff
@@ -34,10 +36,111 @@ class NCVaccine(TableauDashboard):
...
-        df.location_name = df.location_name.str.replace(
-            " County", ""
-        ).str.strip()  # Strip whitespace
-        df.loc[df["location_name"] == "Mcdowell", "location_name"] = "McDowell"
-        return df
+        timeran = dt.datetime.now()
+
+        age_df = df[0]
+
+        age_df.rename(columns = {
```
tlyon3, line 62: I'd recommend breaking out each demographic into it's own normalize method (eg `_normalize_age`, `_normalize_sex`, etc...), then calling each of them in the main `normalize` method.

tlyon3 (changes requested):

sglyon:
Hi @kxdan just checking in to see if you've had any luck fixing the SQL issues with the location columns?

Would it help to have a brief call? If so, please reach out on slack

can_tools/scrapers/official/NC/nc_vaccine.py
kxdan: So i attempted to leverage those initially, but it didn't have the capability to do subsheet navigation, I found a similar PR leveraging this package so did so here as well

can_tools/scrapers/official/NC/nc_vaccine.py
kxdan: Any thoughts here @tlyon3 on how best I should handle this?

can_tools/scrapers/official/NC/nc_vaccine.py
kxdan: Sounds good to me, let me know when you have input on the location_id issue and i'll fix all these in one go

kxdan:
hey @sglyon it looked from the above that @tlyon3 was looking into that the issue could be with the location_id, was waiting to here back from him, any thoughts you have would be appreciated too :)

can_tools/scrapers/official/NC/nc_vaccine.py
tlyon3: Just rename this line to `location_name = 'North Carolina'`

tlyon3: The issue is still coming up. This is correct though. Must be a bug in the `put` method. @sglyon, do you have any insight as to why this would be failing? 

```
In [26]: d.find_unknown_location_id(engine, df)
Out[26]:
Empty DataFrame
Columns: [vintage, dt, category, measurement, unit, age, race, ethnicity, sex, value, location_name]
Index: []

In [27]: df.location_name.unique()
Out[27]: array(['North Carolina'], dtype=object)
```

sglyon:
Thanks @kxdan I missed that comment! I'll take a look today

sglyon:
Hey team, found the problem(s):

- We needed to change `location_type` class property to `"state"`. This plus the suggestion that @tlyon3  made about spelling of `"North Carolina"`
- There were some missing `age` values in the `bootstrap_data/covid_demographics.csv` file, I added those two at the bottom of that file
- I added one helper method to check on missing demographic (age, sex, race, ethnicity) 4-tuples. You feed this method your normalized dataframe and it will return only rows that have a missing 4-tuple. The next step would be to add those entries to the csv file from the previous point.

can_tools/scrapers/official/NC/nc_vaccine.py
sglyon: see below about `location_type = "state"`

can_tools/scrapers/official/NC/nc_vaccine.py
kxdan: Bump here, if there's a way to do subsheet navigation with those tools I will use them, but when we looked previous there was not

kxdan: Just checked latest main, package is already in Main under TableauScraper==0.1.2, so someone else needed it too, removed all diff on adding new package since it's already in amin

kxdan:
Should be good to go in, tests pass local

sglyon:
Data is looking much better, thank you @kxdan!

A few comments about changes I made:

- I split the different demographic group scrapers into separate classes
- I added a couple more class attributes that make it possible to write a general `fetch` and `normalize` method once and get it "for free" for the other demographic groups.

A few more comments about the ideal data to be scraping:

- Could we get county level data from this scraper? I believe we can by setting some tableau workbook parameters.
- The (total_vaccine_initiated, new, percentage) triplet is unfortuantely what the dashboard shows by default, but I don't think it is the most helpful way to look at this data. A preferred triplet would be (total_vaccine_initiated, current, percentage) for total percentage of people that have at least one dose. or even better (total_vaccine_initiated, cumulative, people) for a count of total people with at least one dose. This is another place where we will need to set tableau dashboard parameters and re-query the data
- We'll also need to get total_vaccine_completed data from this scraper. Again I think this can be obtained by setting a Tableau Dashboard parameter of filter

Is this something you can work towards? I'll cc @smcclure17, who has recently become and expert in working with parameterized tableau dashboards

smcclure17:
Updated to track (`total_vaccine_initiated/completed`, `current`, `percentage`) by county. I could not find any cumulative values, or any population values to use to create cumulative values.

smcclure17:
re-added the county data scraper (NCVaccine). 

Added filter function to ensure that scrapers pull from "All sources" data (not just NC sources) 

We might want to update how we specify filter functions (maybe pass a dictionary of `{Name:Value}` keys/val pairs instead of individual strings), but for now I just added a third filter function variable.

sglyon:
Thanks @smcclure17 I tried running this and it took a *very* long time (why on earth does NC have 100 counties?!?!?)

I think it might be worth parallelizing the requests. What do you think?

smcclure17:
yeah it does take absolutely forever--that's smart, I can look into it. 

(and I'm sure no one would notice if they combined half the counties together, what's the harm? 😭😭)

smcclure17:
**Time-series v.s Most Recent Data:**

There's a time-series table (which is being used now) that returns data running back to December, and there's another table that just provides the most recently reported data. The time-series data seems to lag between 2-7 days behind the other table (atm, the time-series data was last updated on 5/24/21, and the second table on 5/28). So, using the most-recent-only table could give us more up-to-date data, but we would lose the historical data. 

Do we have a preferred option? Should we use the time-series data even if it is a few days/a week behind? Should we use only the most recent data? Or, should we collect the time-series data once, and then use only the newest data going forward (as the historical data shouldn't change)?

Any thoughts/suggestions? @sglyon @BrettBoval @mikelehen 

---

Also, for a couple of counties, the time-series values are quite different (lower) than the most-recent table (on the dashboard as well). I only noticed this in two counties (I checked about ~35-40 in total). I'm not sure if this is a reporting lag or if the numbers are incorrect, but the differences in these counties are _much_ greater than in the other counties. For the other counties, the time-series data is behind by at most ~1% (due to the reporting lag I assume), but these counties are off by upwards of 10%. 

The data in the bar-charts are from the "most-recent" table, and the time-series data are from the time-series table. As seen in both, the scraped time-series data matches the dashboard, but it is well below what has been most recently reported.

image
image


image
image

mikelehen:
I don't have a super strong opinion, but here's my 2 cents:
* In general I think this data is going to be used for retrospective analysis rather than day-by-day monitoring so I'd prioritize historical data over realtime updating. This also has the benefit that it may be improved over time (e.g. errors found / fixed retrospectively).  Being a few days (or a week) lagged doesn't seem like a big issue.
* That said, it's a pretty big bummer that the historical data seems to have errors (Gates county is at 28% fully vaccinated overall, so the historical data seems incorrect).

So I'm on the fence but I think I'd be inclined to just use the historical data since in theory it's the more valuable data source (and maybe the errors will get fixed later).  If we could find a way to report the errors to NCDHS, that would be nice.

That said, I definitely wouldn't be opposed to using the non-historical data (with or without an optional one-time import of the historical data) if anybody was keen to do that instead.

sglyon:
I'd also vote for using the time seres

smcclure17:
Ok great, it seems like the time series is the way to go, and if so then I think we should be ready to merge. 

@sglyon Any final thoughts?

smcclure17:
looks like they now surface counts, as well as percentages. I'll switch this to collect counts instead of %s and review/merge by Mon.

PR #221, opened 2021-04-06
Wy vaccine patch2

blew2015:
Updated JSONs being passed to the Google Data Studio API to prevent the job from failing. 

@sglyon @tlyon3


tlyon3 (approved):

sglyon:
Thanks @blew2015 and @tlyon3

PR #222, opened 2021-04-08
FEAT: Alabama Demographics

tlyon3:
Adds scrapers for `total_vaccine_doses_administered` for alabama by race, age, and sex

### DIFF
--- a/can_tools/scrapers/official/AL/al_vaccine.py
+++ b/can_tools/scrapers/official/AL/al_vaccine.py
@@ -34,3 +34,44 @@ class ALCountyVaccine(ArcGIS):
         locations_to_drop = [0, 99999]
         data = data.query("location != @locations_to_drop")
         return data
+
+
+class ALCountyVaccineSex(ALCountyVaccine):
+    variables = {
+        "F": CMU(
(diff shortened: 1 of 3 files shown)


can_tools/scrapers/official/AL/al_vaccine.py
```diff
@@ -34,3 +34,133 @@ def normalize(self, data):
...
+            unit="doses",
+            sex="unknown",
+        ),
+    }
+
+    variable_columns = ["RECIP_SEX"]
+    sheet_num = 6
+
+    def fetch(self):
+        service = "Vaccination_Dashboard_AGOL_v4_PUBLIC_VIEW"
```
sglyon, line 65: looks like we can make this a class attribute also, then we wouldn't need to define a `fetch` method in the subclasses

sglyon (changes requested):

can_tools/scrapers/official/AL/al_vaccine.py
tlyon3: done

sglyon (approved):

PR #225, opened 2021-04-08
FEAT: Lousianna demographics

tlyon3:
Adds county level scrapers for race, age, and sex.

NOTE: The raw data reports in percentages. They provide a population column inline with the data. The values were calculated by taking `floor(percentage_[demographic] * population)`


can_tools/scrapers/official/LA/la_county.py
```diff
@@ -48,3 +48,226 @@ def normalize(self, data):
...
+            measurement="cumulative",
+            unit="people",
+            sex="unknown",
+        ),
+    }
+
+    def normalize(self, data):
+        df = self.arcgis_jsons_to_df(data)
+        # Multiply each of these columns by the population column
+        init_col_names = [
```
sglyon, line 234:
```
init_col_names = [x.replace("_value", "") for x in self.variables.keys())
```

sglyon (changes requested):

sglyon:
great work!

PR #226, opened 2021-04-08
Create main flow that runs all scrapers

ghop02:
This creates a single flow that runs all of the scrapers - I think it'll make it easier to see when data was updated and what scrapers are failing instead of having to rely on prefect failed task uis.  Also, it gives a bit more of a sequence: run all scrapers -> update parquet file. 

Right now i dont' think this parallelized.  My thinking is that it doesn't matter too much, but if the total run takes more than 3 hours I think we can look into fixing that

### DIFF
--- a/services/prefect/flows/generated_flows.py
+++ b/services/prefect/flows/generated_flows.py
@@ -1,5 +1,5 @@
+from typing import Any, Tuple, Type, List
 from datetime import timedelta
-from typing import Any, Tuple, Type
 import sentry_sdk
 import pandas as pd
 import sqlalchemy as sa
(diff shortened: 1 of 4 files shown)


services/prefect/flows/generated_flows.py
```diff
@@ -133,18 +137,45 @@ def create_cdc_all_states_flow():
...
+                flow_name=flow.name, project_name=project_name, wait=True
+            )
+            tasks.append(task)
+
+        parquet_flow = StartFlowRun(
+            flow_name="UpdateParquetFiles", project_name=project_name, wait=True
+        )
+
+        for task in tasks:
+            task.set_downstream(parquet_flow)
```
sglyon, line 156: Does this line mean parquet will only run if all scraper flows succeed, or does that adjustment you make to the parquet flow function `trigger=triggers.all_finished` make it so it just waits for all to _finish_ success or not?

ghop02: yup! that's correct - will wait to trigger that until all the upstream tasks are finished (not based on success or failure)

sglyon: Let's give it a shot ;)

sglyon (approved):

PR #231, opened 2021-04-12
FEAT: Michigan Demographics

tlyon3:
This is almost ready to go. I can't seem to insert into the test database because of an unknown variable ID. I've checked the spelling and made sure the `age` and `sex` combinations are in the `covid_demographics` files, but I still can't push.

@smcclure17, could you take a look and try to find it? If you can't push it to @sglyon


smcclure17:
tests were failing because the what we wanted as the `category` column was labeled as `variable`. And, each CMU pair/variable has to exist in the `covid_variables.csv` file, so I added those. 

I believe Michigan reports Detroit as it's own county,  even though it is a part of Wayne county. So, I think the Detroit and Wayne county entries have to be combined to get accurate #'s for Wayne county.

smcclure17 (commented):
see comment, but looks good!

PR #233, opened 2021-04-12
County-level vaccine scraper for ND

MattSidor:
First dose and "up-to-date" (fully vaccinated).

Their Microsoft BI dashboard is here: https://www.health.nd.gov/covid19vaccine/dashboard

Dashboard only provided percentages of coverage, so I pulled population counts from their COVID infection rate dashboard (https://www.health.nd.gov/diseases-conditions/coronavirus/north-dakota-coronavirus-cases) to multiply the percentages by.


MattSidor:
Tagging @tlyon3 and @sglyon for this. Thank you.

can_tools/scrapers/official/ND/common.py
```diff
@@ -0,0 +1,60 @@
+# ND County Populations
+# Retrieved from ND COVID Case Count Dashboard
+# https://www.health.nd.gov/diseases-conditions/coronavirus/north-dakota-coronavirus-cases
+# Based on 2018 counts
+
+nd_populations = {
```
tlyon3, line 6: Are we worried about population fluctuation?

sglyon: I don't think so. For me this is an acceptable way to move from percent to counts

tlyon3 (approved):

PR #234, opened 2021-04-13
WV: Updated query to match new variable names from WV's API

MattSidor:
It looks like they changed the names of some of their variables and they moved first/both doses into the same table, 'COVID_Vaccine'. I think this fixes it.


MattSidor:
@sglyon 🙏

tlyon3 (approved):
Looks good. Thanks

sglyon (approved):

PR #235, opened 2021-04-13
FEAT: Virginia demographics

tlyon3:


can_tools/scrapers/official/VA/va_vaccine.py
```diff
@@ -81,3 +78,288 @@ def normalize(self, data) -> pd.DataFrame:
...
+
+class VirginiaCountyVaccineDemographics(VirginiaVaccine):
+    viewPath = "VirginiaCOVID-19Dashboard-VaccineDemographics/VirginiaCOVID-19VaccineDemographics"
+    filterFunctionName = (
+        "[federated.1a55gj51go17vw1cds8i50u4b8by].[none:Locality Name:nk]"
+    )
+    filterFunctionValue = None
+    has_location = False
+    location_type = "county"
+    counties = [
```
sglyon, line 91: fly by comment.... Seems like we should be able to pull this from the CSV files in bootstrap_data instead of listing them out here. Would it be worth the effort to do that?

can_tools/scrapers/official/VA/va_vaccine.py
```diff
@@ -81,3 +78,288 @@ def normalize(self, data) -> pd.DataFrame:
...
+        df.category = df.category.str.replace(
+            "People Fully Vaccinated", "total_vaccine_completed"
+        )
+        df.category = df.category.str.replace(
+            "People with At Least One Dose", "total_vaccine_initiated"
+        )
+
+        df.race = df.race.str.replace("White", "white")
+        df.race = df.race.str.replace("Other Race", "other")
+        df.race = df.race.str.replace("Native American", "native_american")
```
smcclure17, line 333: It looks like we have both `ai_an` (American Indian or Alaska Native) and `native_american` in our demographics schema. Is this intended? I've seen Native American categorized as `ai_an` in the past (in our scrapers), but I don't know if there's a distinction.

can_tools/scrapers/official/VA/va_vaccine.py
smcclure17: This should/might work for this: 
```
import os.path
counties = list(
            pd.read_csv(
                os.path.dirname(__file__) + "/../../../bootstrap_data/locations.csv"
            ).query("state == @self.state_fips and name != 'Virginia'")["name"]
        )
```

smcclure17 (commented):
I think this looks good! 🙌

can_tools/scrapers/official/VA/va_vaccine.py
tlyon3: I put it as just Native American because that's how the data had it...but maybe just putting our existing `ai_an` will be better. I'll change it

can_tools/scrapers/official/VA/va_vaccine.py
tlyon3: @sglyon, does the same file structure exist on the server? If so, then @smcclure17 code is good.

can_tools/scrapers/official/VA/va_vaccine.py
smcclure17: That's fair--I ask in large part because I'm not sure myself

smcclure17 (approved):

sglyon (approved):

PR #236, opened 2021-04-14
Update South Dakota Vaccine Scraper

smcclure17:
Updates the SDVaccineCounty scraper to collect `total_vaccine_initiated`,`total_vaccine_completed`, and `total_vaccine_doses_administered`, and first/second dose data by manufacturer.

I could not find a neat solution to #199, in that I still don't know why the request limits the number of responses to 12 or 13 counties. To circumvent this issue I broke up the fetch into 6 separate requests, returning data for 11 counties in each. Then I combined the results. If anyone knows of a fix/a more efficient way to approach this please let me know, but this should work correctly.

One question: If we collect manufacturer/dose specific data, do we want to collect/return that data? Or do we only care about the standard initiated/completed/administered?

### DIFF
--- a/can_tools/bootstrap_data/covid_variables.csv
+++ b/can_tools/bootstrap_data/covid_variables.csv
@@ -180,6 +180,7 @@ total_vaccine_initiated,new,people
 total_vaccine_initiated,new,doses
 total_vaccine_initiated,new_7_day,people
 pfizer_vaccine_completed,cumulative,people
+janssen_vaccine_completed,cumulative,people
 total_vaccine_allocated,cumulative,doses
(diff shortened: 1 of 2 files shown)


tlyon3 (changes requested):
If the manufacturer specific data (ex `moderna_vaccine_initiated/completed`) is available, I beleive we want that. That data is more granular and end users (or ourselves) can aggregate later down the line if wanted/needed.

smcclure17:
great, thanks @tlyon3. 

Added trackers for each dose

tlyon3 (approved):

sglyon:
Great work!

PR #238, opened 2021-04-15
Update DC Vaccine demographics

smcclure17:
~~This still needs a long-term fix for when DC changes the order of the data on their dashboard, but this updates it temporarily~~

_see most recent commit + comment_

### DIFF
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -21,11 +21,8 @@ from can_tools.scrapers.official.DC.dc_deaths import DCDeaths
 from can_tools.scrapers.official.DE.de_vaccine import DelawareCountyVaccine
 
 from can_tools.scrapers.official.DC.dc_vaccines import (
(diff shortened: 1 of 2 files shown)


tlyon3 (approved):

sglyon (changes requested):
Seems to be broken again... :(

The `Sex` scraper seems to be gathering data for race

The race scraper is broken

Age and ethnicity seem correct

smcclure17:
This last commit should fix the ongoing headache these scrapers have been. Removes the individual demographic classes and creates `DCVaccineDemographics`. All demographic data is fetched at once using the `TableauScraper` module, and handled in one `normalize` function. This should eliminate the issue of the dashboard changing the ordering of the results.

tlyon3 (approved):

sglyon (approved):

sglyon:
Thank you for fixing this @smcclure17!

PR #243, opened 2021-04-20
WIP: update Wisconsin vaccines

smcclure17:
This fixes the scraper, returning it to its previous functional state. 

### **Issues**: 
* The `total_vaccine_completed` (`Completed Series (#)`) values returned in the data **_do not_** match those surfaced on the dashboard. They appear to always under-represented (screenshots below), which leads me to believe that Johnson and Johnson series might not be included in the raw data. J&J doses are reportedly included in the vaccine completed numbers on the dashboard (see definition in the screenshot). I cannot find any manufacturer-specific data, so I'm unsure how to verify/investigate further. With this in mind, I'm not sure how to proceed.

* ~~Currently (and previously), the demographic scrapers only collect(ed) state-level data, however, there appears to be county-level data available. This would require some filtering/new requests/digging to access. Is it a priority to collect county-level demographic data (meaning: should I make an effort to do so)? I'm more than willing to, but I want to make the best use of time.~~

### To be added: 
* County-level demographic data

---
example of mis-matching `total_vaccine_completed` values:

image
image

definition of series complete data:
image

### DIFF
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -128,11 +128,14 @@ from can_tools.scrapers.official.VT.vt_vaccinations import (
 from can_tools.scrapers.official.WA.wa_vaccine import WashingtonVaccine
 from can_tools.scrapers.official.WI.wi_county_vaccine import WisconsinVaccineCounty
 from can_tools.scrapers.official.WI.wi_demographic_vaccine import (
-    WisconsinVaccineAge,
-    WisconsinVaccineEthnicity,
-    WisconsinVaccineRace,
-    WisconsinVaccineSex,
-    WisconsinVaccineCountyDemographics
+    WisconsinVaccineStateAge,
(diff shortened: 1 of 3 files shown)


tlyon3 (approved):

sglyon:
@smcclure17  is this sill considered a draft?

Thanks!

smcclure17:
It's a draft in that it still has the mismatching dose data and only collects state-level demographics. But, if we want to merge in the meantime while those are worked on, then this is good to go.

sglyon:
Let's wait until we finish all the outstanding items. Thanks!

smcclure17:
added demographics by county. This is very slow (2-3 min per scraper) because 2 calls have to be made per county (one for initiated values and one for completed). If wanted, I could look to restructure the fetch to try and collect all the demographics at once/combine the scrapers into one.

PR #244, opened 2021-04-20
FIX: Update service name for ALCountyVaccine

tlyon3:
The general `ALCountyVaccine` class was using an old ArcGIS resource causing flat data since March 26. 

reference: https://trello.com/c/D1xpkreq/81-alabama-county-vaccine-data-is-flat-since-march-26


tlyon3:
Trello Card: https://trello.com/c/D1xpkreq/81-alabama-county-vaccine-data-is-flat-since-march-26

sglyon:
Hey @tlyon3 thank you for taking the initiative on this

I actually fixed it yesterday: https://github.com/covid-projections/can-scrapers/commit/a799f466c28dceddc5185202e7e804a311bff4d0

I will make sure to communicate things better so we don't duplicate efforts going forward

PR #245, opened 2021-04-20
FIX: Philadelphia County Vaccine

tlyon3:
I believe the fix introduced in #239 was getting overwritten by the state scraper. This removes the rows the `PennsylvaniaCountyVaccines` scraper gets for Philadelphia county as they are handled by the `PhiladelphaVaccine` scraper.


sglyon:
Let’s leave the data from the state dashboards. Might be helpful as a comparison tool

tlyon3:
I left the data and changed the superclass in the `PhiladelphiaVaccine` scraper

sglyon:
Ahh my fault, I see that now.

Thanks

PR #248, opened 2021-04-24
Add Ohio county vaccine demographics

smcclure17:
I couldn't find a solution to make multiple `.getParameter()` calls on the same `TableauScraper` instance, so this follows the same structure as the `WisconsinVaccineCounty...()` demographic scrapers. 

gets data for (`total_vaccine_initiated`, `cumulative`, `people`) and (`total_vaccine_completed`, `cumulative`, `people`)

### DIFF
--- a/can_tools/scrapers/official/OH/oh_vaccine_demographics.py
+++ b/can_tools/scrapers/official/OH/oh_vaccine_demographics.py
@@ -51,8 +51,6 @@ class OHVaccineCountyRace(StateDashboard):
         ws = ts.getWorksheet("New Map")
         counties = ws.getSelectableValues("county")
         for county in counties:
-            if county == "Washington":
-                break
             print("working on county: ", county)
             wb = ws.select("county", county)
             df = wb.getWorksheet("New Demographics").data.assign(location_name=county)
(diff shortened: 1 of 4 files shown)


tlyon3 (approved):

can_tools/scrapers/official/OH/oh_vaccine_demographics.py
```diff
@@ -1,22 +1,123 @@
...
+        )
+        df = df[
+            ["location_name", "Selected Demographic-alias", "initiated", "complete"]
+        ]
+        df["Selected Demographic-alias"] = df["Selected Demographic-alias"].str.lower()
+        if self.rename_key:
+            df = df.replace(self.rename_key)
+
+        out = self._reshape_variables(df, self.variables)
+        out = (
```
sglyon, line 83: looks like this might be repeating some of the work done in `._reshape_variables`?

In particular this code does `out = self._reshape_variables(df, self.variables)` and then overrides `out` when we do `out =(df.melt)` -- i.e. the first `out` isn't used.

sglyon: I see what you did -- I will just delete the `_reshape_variables` call as it is not needed

sglyon (approved):

can_tools/scrapers/official/OH/oh_vaccine_demographics.py
smcclure17: whoops--thanks for the catch. I was trying to see if I could use `._reshape_variables` with demographic columns then must have forgotten to delete it afterwards

PR #249, opened 2021-04-25
add South Carolina county vaccine demographics scraper

smcclure17:
adds a scraper to collect vaccine demographic data for South Carolina. 

They appear to only collect/store `total_vaccine_initiated` data--that's all I found surfaced on the dashboard, and I did not see any potential filters/parameters to find vaccine completed data in their tables. 

Data is collected for combinations of race, sex, and age demographics. 

There are data values coded as `-1`, which I assumed (I could not find a definition anywhere) to mean NA/non-existent, so I removed them. If there's a better/correct way to handle this we can change it.


sglyon:
yeah I think for now we will leave this as WIP until they get all the data.

The issue we would face is that we would have 1st dose from this dashboard and then 2nd dose from the CDC

We've roughly decided (as a larger CAN team) that mixing reporting sources for each dose number is problematic.

Thanks for working on it @smcclure17

smcclure17:
that makes sense--sounds good!

PR #250, opened 2021-04-26
Added demographic data for Maine

blew2015:
One of the tests for the MaineAgeVaccines fails for me because Maine is saying more people aged 60-69 have completed their vaccination than have started their vaccinations. I wonder if this is due to the J&J vaccine only needing one dose, and therefore showing up exclusively in the total_vaccine_completed column. @sglyon @tlyon3 let me know your thoughts for how we want to deal with this. I am open to any and all ideas. 

I also had to tweak the demographics csv to allow me to give 0-15 as an age range. As always let me know if there are issues.


sglyon:
Thank you for adding things

- adding a row to the csv file was perfect -- thank you
- Unfortunately the data reported on the dashboard is inconsistent with how we are reporting data in CAN. We want 1st dose to be "people with at least one dose of any vaccine" and completed to be "people with a full series of any vaccine". Based on the descriptions below the dashboard on their site, the completed variable matches what CAN wants, but the 1st dose only includes `moderna_1 + pfeizer_1`. As you noted, this is missing the J&J doses


Do you know if they give any more info somewhere inside the query responses that just don't make their way into the visualizations?

blew2015:
I will look into this tonight.

blew2015:
Thanks Sean! Not sure why I didn't see this in your code earlier. Thanks for your help. Ill put these changes in tonight.

smcclure17:
np at all! Also, we might be able to get county-level data by adding a county table/selection like `MaineCountyVaccine()` does, but idk if this is a priority

blew2015:
One thing I wasn't sure how to deal with: there seems to be a couple instances of second doses being given that were not assigned to a vaccine manufacturer that are not being accounted for in the current code I have. I would just add 
`total_vaccine_completed`/`total_vaccine_initiated` columns that account for these missing doses but the value given for the first dose would be inconsistent with our definitions (as noted earlier in this thread). I could always add the J&J total to the `total_vaccine_initiated` column if we wanted. I could also just include a `total_vaccine_completed` column and not include the `total_vaccine_initiated`.  These were just a couple of ideas I came up with for how to deal with this, but I wanted to ask this group of people before pushing any of these changes since you guys are much more familiar with the CAN dataset.

sglyon:
Thanks @blew2015 

What would it look like if we were to try your first suggestion and 

>  add the J&J total to the total_vaccine_initiated column

That sounds right to me, but I want to make sure I understand the subtleties

blew2015:
@sglyon [https://github.com/blew2015/can-scrapers/blob/e406f27a54304b8ca327cd0bbca5e72ef0651fae/can_tools/scrapers/official/ME/me_vaccines.py#L324](url) See the linked line for an example of the solution you referenced in your previous comment. 

`record["M"][0]["DM2"][0]["C"][0]` is equivalent to the total number of 2 dose vaccines that have been started (Moderna/Pfizer). That `j_j_vacc` variable is the amount of J&J vaccine that has been given out.

sglyon:
That sounds good to me. 

Is this PR ready for a review?

blew2015:
I believe so!

sglyon:
The data looks great, thanks @blew2015 

Would you be willing to work on extending this to collect county level data also? As noted by @smcclure17, I think this will require making an additional request per county, but it is worth it! @smcclure17 would you collaborate with @blew2015 to get this in?

Thanks!

smcclure17:
yep, I'm free all today + tomorrow to meet/collab! and,  #254 has code that should be pretty applicable for this (although some issues are still being worked out) that might help be helpful to look at

blew2015:
I can certainly get that done. But it won’t be until monday at the earliest. I have some commitments this weekend. 

@smcclure17 i feel pretty confident I can do this but if I have questions can I reach out to you over slack?

smcclure17:
yeah that sounds great!

sglyon:
Great, thanks team let me know when that is ready!

sglyon:
ping! How are we doing on this?

PR #251, opened 2021-04-26
fix philly

sglyon:


mikelehen (approved):
Thanks. Please delete historical data as well!

sglyon:
deleted

PR #252, opened 2021-04-26
fix pa scrapers

sglyon:
- bring in updates from NC pr
- fix pa-scrapers

### DIFF
--- a/can_tools/scrapers/base.py
+++ b/can_tools/scrapers/base.py
@@ -152,6 +152,7 @@ class DatasetBase(ABC):
     data_type: str = "general"
     table: Type[Base]
     location_type: Optional[str]
+    base_path: Path
     source: str
 
     # this list of tuples specifies categories for which the first
@@ -256,6 +257,7 @@ class DatasetBase(ABC):
             "sex",
         ],
         var_name: str = "variable",
+        skip_columns: List[str] = [],
     ) -> pd.DataFrame:
         """
         Adds columns "category", "measurement", and "unit" to df
@@ -274,6 +276,9 @@ class DatasetBase(ABC):
         var_name: str
             The name of the column in `df` that should be used to lookup
             items from the `cmu` dict for unpacking columns
+        skip_columns: List[str] (default=[])
+            Can be set instead of ``columns`` if there are a small number
+            of columns that should not be set
 
         Returns
         -------
@@ -284,6 +289,8 @@ class DatasetBase(ABC):
         variable_column = df[var_name]
         out = df.copy()
         for col in columns:
+            if col in skip_columns:
+                continue
             out[col] = variable_column.map(lambda x: cmu[x].__getattribute__(col))
         return out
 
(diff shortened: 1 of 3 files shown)


sglyon:
all PA scrapers (standard vaccine one and 4 demographic ones) were broken. This fixes them

can_tools/scrapers/base.py
```diff
@@ -274,6 +276,9 @@ def extract_CMU(
         var_name: str
             The name of the column in `df` that should be used to lookup
             items from the `cmu` dict for unpacking columns
+        skip_columns: List[str] (default=[])
+            Can be set instead of ``columns`` if there are a small number
+            of columns that should not be set
```
mikelehen, line 281: Can you clarify what the contract is between this and `columns`?  The comment suggests you can't use them together, but the code looks like they're used together (you'd only add a column to `skip_columns` if it was already in `columns`?)...

can_tools/scrapers/official/PA/pa_vaccines.py
```diff
@@ -124,7 +124,7 @@ def normalize(self, resjson):
 
         # Dump records into a DataFrame
         df = pd.DataFrame.from_records(data_rows).dropna()
-        df = df.query("location_name != '' & location_name != 'Out-of-State'")
+        df = df.query("location_name != '' & location_name != 'Out-of-State*'")
```
mikelehen, line 127: Just to confirm, we don't have to worry about the other location names that the demographics scraper does?  https://github.com/covid-projections/can-scrapers/pull/252/files#diff-110778fd367ad022afb439a15950c6a6b7257e77cf95c7d2d86dcd296956bcd0R194

can_tools/scrapers/official/PA/pa_vaccines.py
```diff
@@ -355,62 +322,37 @@ def normalize(self, resjson):
...
             .pivot_table(
-                index=["location_name", self.demographic],
-                columns="variable",
-                values="value",
+                index=["county", self.demographic],
+                columns="coverage",
+                values="count",
+                aggfunc="sum",
             )
+            # .fillna(0)
```
mikelehen, line 331: intentionally left here?

mikelehen (approved):
Unfortunately I don't know the code well enough to do a terribly deep review, but this looks good superficially at least.  😁

can_tools/scrapers/base.py
sglyon: Yeah this is a bit hacky.

I use it when I want to preserve the value of one of these 7 columns (category, measurement, unit, age, race, ethnicity, sex), but set the value of the other 6. In this PR, I leave `columns` as the default and then set `skip_column = ["age"]` to make sure the age column is not overwritten in this function.

A bit wonky, I agree

can_tools/scrapers/official/PA/pa_vaccines.py
sglyon: Correct!

can_tools/scrapers/official/PA/pa_vaccines.py
sglyon: deleted, thanks

PR #254, opened 2021-04-27
add South Dakota vaccine demographics

smcclure17:
WIP due to issue/discussion #256 

collects `total_vaccine_initiated` and `total_vaccine_completed` at the county level for `age`, `race`, `sex`, and `ethnicity`. Individual doses are provided, so they could be collected as well if wanted.


smcclure17:
hi @sglyon. I'm still unsure what the best solution to #256 is, as at the moment a lot variables are dropped because of it. The variables that are returned are accurate and good to go, but I don't feel great using a scraper that will knowingly not report all of the potential data. 

I could give an hour or so today to finding a solution in the JSON parsing today

smcclure17:
restructured the JSON parsing/dataframe creation to make the forward filling manageable. Since where each "R" value is filled from depends on its location within the JSON, I couldn't find a way to use `pd.df.ffill()`, so the forward filling is done manually. 

I've done a pretty hefty amount of validating and all the numbers appear to line up. The totals by dose for each scraper match those on the dashboard, and all the scrapers return the same aggregated values. 

I can provide my testing/validation code if wanted

sglyon:
Great work @smcclure17 !

sglyon (approved):

PR #259, opened 2021-05-03
add race and age county vaccine scrapers for hawaii

smcclure17:
returns `total_vaccine_initiated/completed` by `cumulative people` and `current percentage` for race and age categories

I could not find a (`total_vaccine_completed`, `cumulative`, `people`) measurement in the `Age` data, so I calculated it using the percentages given. We can remove this if it's preferred that we don't do that. 

I believe percentages are calculated with populations of people over 18, as is stated on the homepage of the dashboard, but it is not explicitly mentioned in the Race view.

image


sglyon:
Thanks @smcclure17 your computation to use the population * pct == people looks right to me. I say we go with it.

good work!

PR #260, opened 2021-05-05
update Maricopa County vaccine scraper

smcclure17:
records `total_vaccine_doses_administered` and `total_vaccine_completed` data for Arizona's Maricopa county. This extracts json after scraping the webpage/[datawrapper dashboard](https://datawrapper.dwcdn.net/Y9bAu/22/), which I hope is stable.

They return first dose data but it doesn't include J&J data so I did not collect it. There is also an "unknown" category, which I'm not sure what to do with (I have it removed it at the moment). 

Also, just to note, there are some dips/drop-offs in the historical data (in February):

_total doses administered:_
total_doses

_total vaccine completed_:
completed

_total first dose (although this data is not included in the scraper)_
total_1_dose

This might be corrected data, but I'm not sure, so I wanted to make note of it.

### DIFF
--- a/can_tools/scrapers/official/AZ/counties/maricopa_vaccine.py
+++ b/can_tools/scrapers/official/AZ/counties/maricopa_vaccine.py
@@ -1,66 +1,66 @@
-import pandas as pd
 import requests
 import us
-from bs4 import BeautifulSoup
-
-from can_tools.scrapers.base import CMU
+import re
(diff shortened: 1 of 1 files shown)


sglyon:
Thanks @smcclure17 let's filter out our return data to only return from 2021-03-01 onwards 

Thanks!

PR #263, opened 2021-05-06
Update Oregon county vaccine scraper

smcclure17:
They've moved their tables around it looks like. I've added a filter function to access each county. This tracks `total_vaccine_initiated/completed` by `cumulative`, `people` and `total_vaccine_doses_administered` by `cumulative`, `doses`.

The below value/variable (people initiated) is what I'm tracking as total_vaccine_initiated, and just want to make sure this is correct.
image</del>


sglyon:
LGTM thank you

PR #265, opened 2021-05-07
add race, age, sex demographics for Massachusetts by county

smcclure17:
data is returned by town/city with it's corresponding county. I've summed the values of each town in each respective county to get the county-level data.


sglyon:
great work!

PR #274, opened 2021-05-19
add race, sex, age, ethnicity demographics for Delaware counties

smcclure17:
Collects `total_vaccine_initiated/completed` by cumulative counts and by percentages for each demographic 

<b>Questions:</b>
- Percentages are calculated from populations of people > 16 years old. I've added variables to `covid_demographics.csv` to reflect this, but if this is not how we handle percentage-based data I can change this.
- They report `patient declined to disclose` and `data not reported` for both `race` and `ethnicity`. How should these be handled? My thought was to maybe combine them together to make a general "`unknown`" variable. I have them removed at the moment so that the tests pass, but this needs to be addressed.


sglyon:
Thank you @smcclure17 I think for now we might want to track only the raw counts data. 

I don't want to create any downstream inconsistencies between the percentages are calculated here and how they are computed for other geographies

Do you agree @TomGoBravo and @mikelehen ?

smcclure17:
That makes sense! I can remove them if that's the conclusion

smcclure17:
@sglyon yep, it's now removed. 

I also added code to combine `patient declined to disclose` and `data not reported` into an `unknown` variable (line 242 to end). Is this a good idea? or should we just not track those two variables

sglyon:
I think that's the right idea, thank you!

PR #281, opened 2021-05-25
update HI county vaccine scraper

smcclure17:
updates the scraper to collect `total_vaccine_initiated/completed` data that includes all sources/distributers. 

the `total doses administered` data on the dashboard matches the data from the 'jurisdiction only' data, so I did not collect it.


sglyon:
This is great, thank you @smcclure17 

I will merge this b/c it is giving correct data.

However, do we need to go back and delete old data?

smcclure17:
@sglyon yeah, we will need to remove the old data. 

I'd be willing to look into that (with the help of your new helper functions 🙌), but I'll need to learn how to get access to the db

PR #283, opened 2021-05-27
allow for multiple re-directs in Maricopa Vaccine County

smcclure17:
If a data card requires multiple re-directs to reach the final URL, the scraper currently fails (I was assuming it would only ever re-direct once, but on the new data it requires two.)

This adds the ability to re-direct as many times as needed, by calling the `_get_json()` function until a valid URL is found. 

This should be safe (never infinitely repeat) as if there's a bad URL/landing page, the `_get_url()` method should break, stopping any sort of loop. But, let me know if there are concerns about this.


sglyon:
This is a great change! Thank you

PR #284, opened 2021-05-27
add Maine county demographics [updated]

smcclure17:
tracks `total_doses_initiated/completed` for age, race, sex at the county level. 

`total_vaccine_initiated` is calculated from the data as `moderna/pfizer_first_dose` + `jj_doses`


sglyon:
Hey @blew2015 and @smcclure17 thanks for working on this.

I was reviewing and noticed that for the age scraper the numbers don't seem to line up

Here's what I see on the dashboard:

Screen Shot 2021-05-27 at 6 02 46 PM


And here's what I see from the scraper:

Screen Shot 2021-05-27 at 6 05 24 PM

Notice the 80+ group. The dashboard shows first shot at 12163 and second shot at 12001

The scraper shows first shot at 24272 and second shot at 24904

Is there an error in how I'm viewing the data?

smcclure17:
oof, thank you @sglyon good catch, not sure how I missed that. 

Looks like there are two problems:
* an easy one: I had half of the `demographic_key` keys off by one, so some of the values were being stored in the wrong location/variable (as you saw), I've fixed that.
* some of the counties have a category for age group `0-11` and others don't. This messes with how we identify/grab the data for each group (we identify each variable/group by its location in the list, so having a changing number of list items causes a problem)

I'll work on a fix for the latter first thing tomorrow.

sglyon:
Thanks @smcclure17 I see that the data now matches the dashboard! Nice work.

I did take a closer look at their data definitions though and found this...

> First dose: People who have received the first dose of a COVID-19 vaccine series from Pfizer or Moderna.
>Final dose: These are fully vaccinated individuals, who have either received the second dose of the Pfizer or Moderna vaccine, or one dose of the Johnson & Johnson vaccine.

Looks like the "first dose" number they report will not be consistent with our definition of "people with at least one dose" (missing the J&J people)

@mikelehen @BrettBoval any suggestions on what to do here? Is this usable for us?

smcclure17:
Is this taken care of by line 402? It should add the J&J doses to the 'first dose' that they report on the dashboard

can_tools/scrapers/official/ME/me_vaccines.py
```diff
@@ -185,3 +186,382 @@ def normalize(self, resjson: dict) -> pd.DataFrame:
...
+    variables = {
+        "initiated_total": v.INITIATING_VACCINATIONS_ALL,
+        "complete_total": v.FULLY_VACCINATED_ALL,
+    }
+
+    col_mapping = {
+        "G0": "county",
+        "M_1_DM3_{demo}_M_0_DM4_0_C_0": "first_dose_total",
+        "M_1_DM3_{demo}_M_0_DM4_0_C_1": "complete_total",
+        "M_1_DM3_{demo}_M_1_DM5_0_C_2": "jj_complete",
```
sglyon, line 212: @smcclure17 what is the meaning of `jj_complete` and `jj_init`? Are they always the same?

mikelehen:
If we can't get the correct accounting for vaccine initiated, I would just report vaccine completed, and then our pipeline should use the state-wide ratio to estimate vaccine initiated from the completed, like we do for county data from CDC, etc.

can_tools/scrapers/official/ME/me_vaccines.py
smcclure17: they seem to store the J&J doses in one of two locations. Each manufacturer returns dose-specific data in the JSON. For the pfizer and moderna vaccines, the first/second doses are stored in a list (inside a dict/object) where the first index is the # of first doses and the second index is the # of second doses. For the J&J doses, they seem to put them arbitrarily in either index (either the 'first dose' or 'second dose' index)

Here's what the data looks like when scraped:
image

sglyon: That makes sense -- good sleuthing and figuring this out. I think we are good to merge!

PR #290, opened 2021-06-01
fix MT vaccine completed

smcclure17:
We were pulling from a strictly "2 dose" column that did not include j&j doses. Updated to pull correct `total_vaccine_completed` data and added `total_vaccine_doses_administered` data.

### DIFF
--- a/can_tools/scrapers/official/MT/mt_vaccinations.py
+++ b/can_tools/scrapers/official/MT/mt_vaccinations.py
@@ -1,7 +1,7 @@
 import pandas as pd
 import us
 
-from can_tools.scrapers import CMU
+from can_tools.scrapers import variables as v
 from can_tools.scrapers.official.base import ArcGIS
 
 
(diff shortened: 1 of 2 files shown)


can_tools/scrapers/official/MT/mt_vaccinations.py
```diff
@@ -18,11 +18,16 @@ class MontanaCountyVaccine(ArcGIS):
             measurement="cumulative",
             unit="people",
         ),
-        "Dose_2": CMU(
+        "Fully_Vaxed": CMU(
```
ghop02, line 21: I bet the MT scraper was a bit old, but could use update the values here to use the variable constants?

ghop02 (approved):
LGTM! Do you think we'll be able to rerun this on old data? (I believe there's a command for that) - I can also follow up on that part!

can_tools/scrapers/official/MT/mt_vaccinations.py
smcclure17: @ghop02 yeah I can add the variables!

I don't know much (anything) about re-running for the old data--but I can look into it. Do you know where/how that might be done?

can_tools/scrapers/official/MT/mt_vaccinations.py
```diff
@@ -1,7 +1,7 @@
 import pandas as pd
 import us
 
-from can_tools.scrapers import CMU
+from can_tools.scrapers import variables as v
```
ghop02, line 4: In general we should try to shy away from single letter variables - does `variables` need to be aliased?

smcclure17: ah ok, no it doesn't--was just for ease of typing haha, I can remove that

ghop02: thanks!

smcclure17: should be fixed!

ghop02 (approved):
LGTM! Thanks for hopping on this!

smcclure17:
i'll wait for the tests and then I'll merge

ghop02:
ooo tests take a very long time - we should really figure out a better way to test - it looks like we essentially run the pipeline locally, which isn't super helpful for unit tests.  Could you just run the tests locally for this scraper and confirm they pass? (I'd just like to get it in sooner rather than later)

smcclure17:
yep they pass, i'll go ahead and merge!

Yeah--they do take a loong time. I think it's mainly b/c the demographic scrapers have to make a lot of requests (often one per county) so they can each take anywhere from 2-10 minutes. 

We run each scraper to make sure that all the locations/demographic values/variables are valid (exist within the `bootstrap_data` files) but that does mean it takes a lot of time. I can put some thought into a better way of doing this

PR #295, opened 2021-06-04
updates/fixes for DC vaccines (county and demographics)

smcclure17:
updates the table/collection and adds `total_vaccine_doses_administered` for `DCVaccines`. Removes excess `all` rows from `DCVaccineDemographics`. 

Question about `DCVaccine`: 
They report data on three groups: DC residents who were vaccinated in DC, DC Residents who were vaccinated outside of DC, and non-residents. 

We only want to collect DC residents, but does this include both those vaccinated inside and outside of DC? Or just those vaccinated in DC? At the moment, I am collecting both within/outside resident groups.


sglyon:
Hmm that is a good question.

I think the right answer is to report DC residents, no matter where they were vaccinated.

It sounds like this might be what you are doing? Is that right?

smcclure17:
yep, at the moment the scraper collects all residents regardless if they were vaccinated in our outside of DC

PR #296, opened 2021-06-04
Re-add Florida county vaccine scraper

smcclure17:
re-add Florida county data--I think it got removed when the demographic scraper was added. Collects `total_vaccine_initiated/completed` by `cumulative` and `new` counts.
 
Also, to note: 
it looks like on the current PDF `male` and `sex-unknown` variables are not being tracked. [See here](https://trello.com/c/STi0O7CT/120-floridacountyvaccine-demographics-currently-not-collecting-male-and-sex-unknown-data). I've tightened the pdf scraper's row boundaries to fix this.


ghop02 (approved):
This looks alright to me!

PR #301, opened 2021-06-09
add individual scrapers for King and Pierce counties, WA

smcclure17:
Added backfill to pull data from API so that historical data is not lost when switching to the new collection method. 

- How do we store/hide API keys? I have the CAN API key stored on my local machine to avoid exposing it, but the scrapers won't succeed at the moment (because I did not push the file it's stored in). Since it's a CAN API key, are we concerned if it's publically accessible?

----
collects `total_vaccine_initiated/completed` for both counties. [Washington State dashboard](https://www.doh.wa.gov/Emergencies/COVID19/DataDashboard) is here for comparison. 

### King County, WA
- [dashboard](https://kingcounty.gov/depts/health/covid-19/data/vaccination.aspx)
- difference from state dashboard: 
--  at least one dose: 78k
-- completed: 48k

### Pierce County, WA
- [dashboard](https://kingcounty.gov/depts/health/covid-19/data/vaccination.aspx)
- difference from state dashboard: 
--  at least one dose: 50k
-- completed: 23k

---

There is a third county scraper for WA (Snohomish county), but the discrepancy is much smaller (only 7k for vaccine initiated)


smcclure17:
messed up some merges so just created a new PR

PR #305, opened 2021-06-12
Add task to monitor consistently failing scrapers

ghop02:
This adds a task that queries the prefect API to find scrapers that have either never succeeded or have not succeeded in the last 24 hours. Hopefully this will bring to light consistently failing scrapers


smcclure17 (approved):
looks great to me!

PR #318, opened 2021-06-22
Add scraper for NE cumulative cases

smcclure17:
Steps taken to disaggregate from health regions to county level:
* calculate health region populations (pop'n) by summing the pop'ns of all the counties within each respective region.
* calculate the percentage of the corresponding health region population for each county (`county pop'n` / `health region pop'n`)
* estimate cases per county my multiplying `% of total health region pop'n` * `total health region cases`

Source of data is [here](https://experience.arcgis.com/experience/ece0db09da4d4ca68252c3967aa1e9dd)

---

**Testing:**

The tests below are run on the data _before it is transformed into long-form_ (return at line 74, before the `_reshape_variables` method), so that the health region and proportion of population data are still accessible.  

Where `d = NebraskaCases().fetch_normalize()`,

Ensure that county cases are never larger than the health region they come from (equal is okay, there are regions with one county):
```
len(d.loc[ d['county_cases'] > d['region_cases']]) == 0
```

Ensure that the proportion of total health region population sums to one across all counties within each health region:
```
d.groupby('health_region').sum()[['county_percent_pop']]
```

Ensure that the sum of the county cases within each health region equals the total number of cases for that health region:
```
# sum county cases in each region
sum_cases = d.groupby('health_region').sum()[['county_cases']].to_dict()["county_cases"]
# find each health region and it's total cases
region_cases = d.groupby('health_region').first()['region_cases'].reset_index()
# compare summed county values and total region values side by side
region_cases['sum_county_cases'] = region_cases['health_region'].map(sum_cases)
```

### DIFF
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -116,6 +116,8 @@ from can_tools.scrapers.official.NE.ne_vaccines import (
     NebraskaVaccineAge,
 )
 
+from can_tools.scrapers.official.NE.ne_cases import NebraskaCases
+
 from can_tools.scrapers.official.NJ.nj_vaccine import NewJerseyVaccineCounty
 from can_tools.scrapers.official.NM.nm_vaccine import NewMexicoVaccineCounty
 from can_tools.scrapers.official.NV.nv_vaccines import NevadaCountyVaccines
--- a/can_tools/scrapers/official/NE/ne_cases.py
+++ b/can_tools/scrapers/official/NE/ne_cases.py
@@ -2,7 +2,7 @@ import pandas as pd
 import us
 import requests
 
-from can_tools.scrapers import CMU
+from can_tools.scrapers import variables
 from can_tools.scrapers.official.base import StateDashboard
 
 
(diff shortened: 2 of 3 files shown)


mikelehen (approved):
This looks good at a high-level.  Python isn't my strongest language, but the code looks reasonable to me.  Let's get this merged and then plumb it into the pipeline (we'll have to explicitly include it in our cases sources since normally we just use NYT for cases) and see what comes out in the snapshot.

cc/ @TomGoBravo who may be able to help with the ingesting.

TomGoBravo:
I didn't notice any bugs from reading the code. Please edit the PR description to describe how you tested this code. It'd be nice to have some kind of check that the result is sane instead of entirely depending on the person doing the daily deploy catching problems that don't raise an exception... but maybe hard to create a test that fails for exactly only true problems.

smcclure17:
Good call @TomGoBravo. I've added some code I used to check the results, let me know if there are other/more tests you'd like to see run. I also manually checked some of the values to make sure the math made sense (eg: make sure that a county with ~50% of the health region population has ~50% of the total cases).

TomGoBravo:
[[blocked 5-26 onward] Nebraska stopped reporting cases at county-level.](https://trello.com/c/IEXT5YRn/1423-blocked-5-26-onward-nebraska-stopped-reporting-cases-at-county-level)

PR #324, opened 2021-06-29
add 1+ dose values from county level CDC data

smcclure17:
CDC has updated their data to include 1+ dose data for ~2700 counties, (they provide vaccine completed data for ~2900 counties).


smcclure17:
no longer necessary as CDC now provides time-series data

PR #337, opened 2021-07-15
add zero filtering to CDC2 vaccine scraper

smcclure17:


can_tools/scrapers/official/federal/CDC/cdc_historical_vaccine.py
```diff
@@ -25,4 +25,6 @@ def normalize(self, data: pd.DataFrame) -> pd.DataFrame:
             data, location_column="FIPS", date_column="Date", locations_to_drop=["UNK"]
         )
         out = self._reshape_variables(out, self.variables)
-        return out.query("location != 78020")
+
+        # remove entries with 0's and a location that caused issues (St. John Island, VI)
+        return out.query("location != 78020 and value != 0")
```
ghop02, line 30: So i think this will do the job, but I think ideally we'd want to just drop trailing zeros for locations - for instance, the leading zeros (i.e. when a place hasn't been vaccinated) isn't a problem. 

I think that solution might be a bit more tricky (have to do grouping and some fancier processing), but I think this is a good case for a bit more context. 

I think this is a good example where the "why" is more important than the what, so adding some context that the CDC dataset is incorrectly reporting missing data at the end of timeseries as 0s would be helpful.  Future me might be confused why we're just deciding to drop all zeros.  Also why did St. John Island cause issues? Would be great if the comment explained that!

smcclure17: good points -- I'll update. My fear was that if a county lost and re-picked up data (eg: zeros in the middle of the time-series) we would want to drop those, too. But, I did some looking and I haven't seen that happen, so it probably isn't an issue. If it is, I could look into dropping all zeros except the leading ones

ghop02: yeah I'm fine with the current approach, but these are all great notes to put in the comment! Don't be scared of making the comment too long ([here's a great example](https://github.com/covid-projections/covid-data-model/pull/1150/files) of a nice comment for some pretty simple code)!

PR #340, opened 2021-07-23
scraper fixes and removals

smcclure17:
- CDCCovidDataTracker began returning suppressed values, causing the scraper to break. This removes those entries, I'm not sure if there is a more graceful way to handle this. 

- The FL scrapers were collecting old data (from June) because the PDF they scrape has not been updated. The TN scrapers were collecting cases and deaths data, so I've disabled them. 

Maine:

-  Removed AstraZeneca doses for the time being, because the variability in which they are reported would mean re-structuring how we parse the JSON, which does not seem worth doing at the moment (when the number of doses is only in the double digits). The reporting should stabilize once the vaccine is approved/more widespread, at which point the query condition should be removed. 

SD:
- Fixed the JSON parse-key ordering (the indices of each dose shifted by one in the list that they're returned in).
- Removed the manufacturer-specific doses from the variables, as I don't believe we have any use for them.

### DIFF
--- a/can_tools/scrapers/__init__.py
+++ b/can_tools/scrapers/__init__.py
@@ -55,11 +55,6 @@ from can_tools.scrapers.official.federal.HHS.hhs_state import (
     HHSReportedPatientImpactHospitalCapacityState,
 )
 
-# Scrapers disabled due to source data no longer being updated
-# from can_tools.scrapers.official.FL.fl_vaccine import (
-#     FloridaCountyVaccine,
-#     FloridaCountyVaccineDemographics,
-# )
 from can_tools.scrapers.official.GA.ga_vaccines import (
     GeorgiaCountyVaccine,
     GeorgiaCountyVaccineAge,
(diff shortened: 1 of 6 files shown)


can_tools/scrapers/__init__.py
```diff
@@ -54,10 +54,12 @@
...
-from can_tools.scrapers.official.FL.fl_vaccine import (
-    FloridaCountyVaccine,
-    FloridaCountyVaccineDemographics,
-)
+
+# Scrapers disabled due to source data no longer being updated
+# from can_tools.scrapers.official.FL.fl_vaccine import (
+#     FloridaCountyVaccine,
+#     FloridaCountyVaccineDemographics,
+# )
```
ghop02, line 62: Is there a reason to keep these around instead of deleting them?

ghop02 (approved):

can_tools/scrapers/__init__.py
smcclure17: nope! no real reason--I'll delete them

PR #342, opened 2021-07-27
Automatically fetch WA iframe url, VA demographics column name fix

smcclure17:
use asyncio to bypass WA's no-javascript blocker, so that the iframe URL will no longer have to be manually updated


ghop02 (approved):
lgtm!

PR #344, opened 2021-07-29
modify WA demographic async method

smcclure17:
updates the async  `_get_from_browser()` method to match the syntax of the method in the WA county scraper. The previous syntax did not work on the server (the connection would hang/timeout), so this should resolve that issue.

### DIFF
--- a/can_tools/scrapers/official/WA/wa_vaccine.py
+++ b/can_tools/scrapers/official/WA/wa_vaccine.py
@@ -84,6 +84,10 @@ class WashingtonVaccineCountyRace(MicrosoftBIDashboard):
     }
 
     async def _get_from_browser(self):
+        """use an async function to wait until the javascript has loaded to extract the iframe url.
+
+        The page is protected by a "no-javascript" blocker, so we cannot parse the html directly.
+        """
         async with with_page(headless=True) as page:
             await page.goto(self.source)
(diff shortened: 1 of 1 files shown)


can_tools/scrapers/official/WA/wa_vaccine.py
```diff
@@ -88,18 +88,14 @@ async def _get_from_browser(self):
...
-        iframe = await page.J(sel)
-        iframe = await page.evaluate(" x => x.outerHTML", iframe)
-        await browser.close()
-
-        return {"src": re.findall(r'pbi-resize-src="(.*?)"', iframe)[0]}
+        async with with_page(headless=True) as page:
+            await page.goto(self.source)
+            sel = "#dnn_ctr34282_HtmlModule_lblContent div"
+            iframe_div = await page.waitForSelector(sel)
+            print("found iframe!")
```
mikelehen, line 95: Just to confirm, is leaving this print() in intentional? (fine with me; just want to make sure it wasn't leftover from debugging).

mikelehen (approved):

can_tools/scrapers/official/WA/wa_vaccine.py
smcclure17: ah, unintentional--I'll remove it!

PR #347, opened 2021-08-12
update CDC data tracker to pull from county endpoints

smcclure17:
This re-purposes and replaces the existing functionality that allows the scraper to be run for all states at once to instead fetch all the counties within one state. Each state is then added as a task to a [prefect flow](https://github.com/covid-projections/can-scrapers/blob/main/can_tools/scrapers/official/federal/CDC/cdc_coviddatatracker.py). 

pulled in the `_retrieve_counties()` method from [existing branch](https://github.com/covid-projections/can-scrapers/compare/create_get_counties_method).

### DIFF
--- a/can_tools/scrapers/official/base.py
+++ b/can_tools/scrapers/official/base.py
@@ -17,7 +17,6 @@ import requests
 from bs4 import BeautifulSoup
 from sqlalchemy.engine.base import Engine
 from sqlalchemy.orm.session import sessionmaker
-from us.states import VA
 
 from can_tools.db_util import fast_append_to_sql
 from can_tools.models import (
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
Nice.  That was less disruptive than I expected. 😁   Hopefully it works out!  The pooling Chris suggested sounds like a good idea to me.  If it's non-trivial and you want another set of eyes on it, I can take another pass.

smcclure17:
@mikelehen just added!

and yeah, the existing "all-state" functionality made it easy, (we now lose the ability for that, but we don't ever use it, and I don't see why we would want to in the future)

mikelehen (approved):
LGTM!  We may want to tune the # of processes depending on how long it takes.

PR #348, opened 2021-08-15
udpate Georgia vaccine scrapers to new source

smcclure17:
shift from the now stagnant ArcGIS endpoint to an xlsx source

### DIFF
--- a/can_tools/scrapers/official/GA/ga_vaccines.py
+++ b/can_tools/scrapers/official/GA/ga_vaccines.py
@@ -1,281 +1,126 @@
 import pandas as pd
+import requests
 import us
 
-from can_tools.scrapers.base import CMU
 from can_tools.scrapers import variables
-from can_tools.scrapers.official.base import ArcGIS
+from can_tools.scrapers.official.base import StateDashboard
 
 
-class GeorgiaCountyVaccine(ArcGIS):
-    """
-    Fetch county level covid data from an ARCGIS dashboard
+class GeorgiaCountyVaccine(StateDashboard):
 
-    Example implementations:
-      * `can_tools/scrapers/official/FL/fl_state.py`
-      * `can_tools/scrapers/official/GA/ga_vaccines.py`
-    """
-
-    ARCGIS_ID = "t7DA9BjRElflTVpw"
     has_location = True
     location_type = "county"
     state_fips = int(us.states.lookup("Georgia").fips)
+    source_name = "Georgia Department of Public Health"
     source = (
         "https://experience.arcgis.com/experience/3d8eea39f5c1443db1743a4cb8948a9c/"
     )
-    source_name = "Georgia Department of Public Health"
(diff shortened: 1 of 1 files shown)


can_tools/scrapers/official/GA/ga_vaccines.py
```diff
@@ -1,281 +1,133 @@
...
+    def normalize(self, data: requests.models.Response) -> pd.DataFrame:
+        sheet = pd.read_excel(data.content, sheet_name=self.sheet_name)
+        data = self._rename_or_add_date_and_location(
+            data=sheet,
+            location_column=self.location_column,
+            # Remove unwanted fips codes
+            # 0 = Georgia
+            # 99999 = Unknown
+            locations_to_drop=[0, 99999],
+            timezone="US/Eastern",
```
mikelehen, line 87: Out of curiosity, why do we need this here but not in the non-demographic scraper?

can_tools/scrapers/official/GA/ga_vaccines.py
```diff
@@ -1,281 +1,133 @@
...
         )
-        df = self._rename_or_add_date_and_location(
-            df, location_column="COUNTYFIPS", timezone="US/Eastern"
+        return (
+            data.pipe(
```
mikelehen, line 90: Mildly curious why you used `.pipe` instead of calling `_reshape_variables` directly here.  It would have made more sense to me if you chained the `pipe` off of the `_reanme_or_add_date_and_location()` return value.  Anyway, not a big deal. 😁

mikelehen (commented):
Thanks for tackling this (and even migrating the demographic scraper)! 🙏

can_tools/scrapers/official/GA/ga_vaccines.py
smcclure17: Good point, there's no reason to use `pipe` here as far as I can tell, so I'm not sure why I did haha. I can update it if it's worth doing!

can_tools/scrapers/official/GA/ga_vaccines.py
smcclure17: The demographic sheets don't have date columns (they aren't time series) so we default to using the current date. Here, we provide `_rename_or_add_date_and_location` with the timezone and the method creates a new `dt` column (with the corresponding timezone). The non-demographic data has a date column, so we instead use the method to convert the existing column to date format. 

afaik, we don't do any conversion/standardization of timezones, so we only use the `timezone` parameter to create the date column.

can_tools/scrapers/official/GA/ga_vaccines.py
mikelehen: Heh, nah, definitely not worth doing. 😁

can_tools/scrapers/official/GA/ga_vaccines.py
mikelehen: Ahhh, gotcha'.  Thanks for explaining.

PR #352, opened 2021-09-13
Update IL vaccine source, fix ND and Maricopa County, AZ  scrapers

smcclure17:
* removes empty entries in Maricopa scraper

* ND restructured their PowerBI tables, this updates the scraper to match

* Updates the IL scraper to [pull from the demographic data](https://trello.com/c/r6L4YOX9/1496-scrape-il-demographic-data-to-get-better-1-dose-data) to get 1+ dose date

### DIFF
--- a/can_tools/scrapers/official/IL/il_vaccine.py
+++ b/can_tools/scrapers/official/IL/il_vaccine.py
@@ -44,6 +44,7 @@ class IllinoisVaccineCounty(StateDashboard):
             .pipe(
                 self._rename_or_add_date_and_location,
                 location_name_column="CountyName",
+                location_names_to_drop=["Illinois"],
                 location_names_to_replace={
                     "Dekalb": "DeKalb",
                     "Dupage": "DuPage",
--- a/can_tools/scrapers/official/IL/il_vaccine.py
+++ b/can_tools/scrapers/official/IL/il_vaccine.py
@@ -14,6 +14,9 @@ class IllinoisVaccineCounty(StateDashboard):
     location_type = "county"
 
     variables = {
+        # AdministeredCount appears to be the column for 1+ doses
+        # and TotalAdministered is the column for total doses administered
+        # AdministeredCount is always greater PersonsFullyVaccinated and less than TotalAdministered
         "AdministeredCount": variables.INITIATING_VACCINATIONS_ALL,
         "PersonsFullyVaccinated": variables.FULLY_VACCINATED_ALL,
     }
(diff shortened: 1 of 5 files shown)


can_tools/scrapers/official/IL/il_vaccine.py
```diff
@@ -11,88 +10,52 @@ class IllinoisVaccineCounty(StateDashboard):
...
+    url = "https://idph.illinois.gov/DPHPublicInformation/api/COVIDVaccine/getCOVIDVaccineAdministrationCountyAge"
     location_type = "county"
 
-    def fetch(self) -> dict:
-        res = requests.get(self.url)
-        if not res.ok:
-            msg = f"Could not request data from {self.url}"
-            raise ValueError(msg)
+    variables = {
+        "AdministeredCount": variables.INITIATING_VACCINATIONS_ALL,
```
mikelehen, line 17: Are we sure this is correct?  "AdministeredCount" sounds like it'd be TOTAL_DOSES_ADMINISTERED_ALL, no?  If we're sure this is right, perhaps add a comment?

can_tools/scrapers/official/IL/il_vaccine.py
```diff
@@ -11,88 +10,52 @@ class IllinoisVaccineCounty(StateDashboard):
...
+                ],
+            ]
+            .pipe(
+                self._rename_or_add_date_and_location,
+                location_name_column="CountyName",
+                location_names_to_drop=[
+                    "Illinois",
+                    "Out Of State",
+                    "Unknown",
+                    "Chicago",
```
mikelehen, line 48: Out of curiosity, Chicago shouldn't exist anymore at this point, right?

mikelehen (approved):
Maybe double check the AdministeredCount mapping, otherwise looks good.

can_tools/scrapers/official/IL/il_vaccine.py
smcclure17: It appears to be, although I don't see a way to truly verify it. There is a `TotalAdministered` column which looks like `total_vaccine_doses_administered`. The `AdministeredCount` seems like it lines up with `total_vaccine_initiated`, as it is never less than the `PersonsFullyVaccinated` column, and roughly lines up with our estimated 1+ dose (although is usually lower, often by a fair margin). 

I can add a comment to clarify this if we feel comfortable with shipping

can_tools/scrapers/official/IL/il_vaccine.py
smcclure17: ah, yeah it will be merged with Cook County, I'll remove this!

can_tools/scrapers/official/IL/il_vaccine.py
mikelehen: 👍   Sounds good.  Let's ship it!

mikelehen (approved):

can_tools/scrapers/official/IL/il_vaccine.py
smcclure17: My mistake, "Illinois" still needed to be removed so I've added just that one location back

PR #354, opened 2021-09-20
add duplicate row removal to _reshape_variables

smcclure17:
some instances of the `CDCCovidDataTracker` were failing on insertion into the database (`put()`) because of duplicated entries in the dataframe. 

There shouldn't ever be duplicated rows in the scraper output data (as far as I can think of), so this adds a line to remove them. `_reshape_variables()` is a helper method used by many scrapers (including `CDCCovidDataTracker`) to format the data

### DIFF
--- a/can_tools/scrapers/official/base.py
+++ b/can_tools/scrapers/official/base.py
@@ -175,7 +175,7 @@ class StateDashboard(DatasetBase, ABC):
         if "vintage" not in data.columns:
             data["vintage"] = self._retrieve_vintage()
 
-        return data
+        return data.drop_duplicates()
 
     def _retrieve_counties(
         self,
--- a/can_tools/scrapers/official/base.py
+++ b/can_tools/scrapers/official/base.py
@@ -139,7 +139,7 @@ class StateDashboard(DatasetBase, ABC):
             Map from column name to output variables
         id_vars: Optional[List[str]], (default=None)
             Variables that should be included as "id_vars" when melting from wide to long
-        drop_duplicates: 
+        drop_duplicates:
             if true duplicate rows are removed, else they are ignored
         kwargs:
             Other kwargs to pass to `self.extract_CMU`
--- a/can_tools/scrapers/official/base.py
+++ b/can_tools/scrapers/official/base.py
@@ -126,6 +126,7 @@ class StateDashboard(DatasetBase, ABC):
         data: pd.DataFrame,
         variable_map: Dict[str, CMU],
         id_vars: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


can_tools/scrapers/official/base.py
```diff
@@ -175,7 +175,7 @@ def _reshape_variables(
         if "vintage" not in data.columns:
             data["vintage"] = self._retrieve_vintage()
 
-        return data
+        return data.drop_duplicates()
```
mikelehen, line 178: Out of curiosity, in this case are there duplicates in the incoming `data`?  Or is it only after reshaping the variables that we end up with duplicates?

Part of me wonders if it's a bit odd for incoming data to have duplicates and maybe we should try to catch it more upstream (i.e. have the CDC Covid Data Tracker scraper remove the duplicates) and optionally have an assert here to _verify_ there are no duplicates (rather than remove them). I don't feel that strongly though, mostly just wondering. 😁 

On the other hand if the duplicates are being "created" through the process of reshaping variables then removing them here makes total sense.  

In any case, I wonder if you could craft a comment explaining why / how the duplicates end up here (or provide an example) so the casual reader like myself doesn't see this line and think, "Hrm, I wonder why there are duplicates!".  But if you don't think it's easy / useful to do, feel free to disregard!

mikelehen (approved):
LGTM.  Optional comment request, but feel free to ignore.

can_tools/scrapers/official/base.py
smcclure17: In this case the duplicates are in the incoming data--to be honest I didn't track it back too far, so I'll look for an explanation. 

I think you're right that it's probably best to not silently remove the duplicates. I'll change this to an assert + add the removal in the CDC Tracker scraper.

mikelehen: Awesome, thanks.  If it ends up being a pain to track down / whatever, you could also take the middle-ground and add a comment like:

> # TODO(smcclure17): Not 100% sure why we need this drop_duplicates() but the CDC Covid Data Tracker scraper is ending up with duplicate rows for some states.

or something along those lines. 😁

smcclure17: There seem to be rows that are almost duplicated in the original data, then become duplicates when we drop other columns. It only seems to happen with rows that are all zeroes, so I'm not super concerned. I'll add a note

image

image

smcclure17: having an `assert` in `_reshape_variables` caused problems because the duplicates were created inside that method, so I instead just added an optional parameter to drop the duplicates if specified

PR #357, opened 2021-09-29
refactor DE county vaccine scraper

smcclure17:
The DE scraper was failing--this PR gets it back online and refactors the scraper. I've removed `total_doses_administered` as we don't have a use for this data, and removing it simplifies things.


mikelehen (approved):
Looks good to me.  Thanks for keeping the scrapers going!

PR #358, opened 2021-09-30
change NJ vax scraper sheet number

smcclure17:
a one-character change haha--NJ changed the table in which they store their vaccine data


mikelehen (approved):
Hah, nice!

PR #359, opened 2021-10-11
scraper fixes and edge case prefect task change

smcclure17:


can_tools/scrapers/official/PA/philadelhpia_vaccine.py
```diff
@@ -21,40 +21,56 @@ class PhiladelphiaVaccine(TableauDashboard):
     source_name = "Philadelphia Department of Public Health"
     baseurl = "https://healthviz.phila.gov/t/PublicHealth/"
```
smcclure17, line 22: the dashboard split first and second doses into two tables, this tries to parse both tables while keeping the same general structure that was used before.

PR #360, opened 2021-10-18
scraper fixes, OH, DE, NC, OH, PHI, MI, SD

smcclure17:
I'm not sure if we ever came to a formal decision, but IIRC our plan was to not track/count booster shots. As such, I've removed booster shots from any scrapers where they caused problems (eg: Ohio and South Dakota). If we think it'd be useful to have, I could update our schema to account for boosters (maybe adding something like `total_vaccine_additonal_dose`)

### DIFF
--- a/can_tools/scrapers/official/NC/nc_vaccine.py
+++ b/can_tools/scrapers/official/NC/nc_vaccine.py
@@ -5,6 +5,7 @@ from can_tools.scrapers import variables
 from can_tools.scrapers.official.base import TableauDashboard
 from tableauscraper import TableauScraper
 
+
 class NCVaccine(TableauDashboard):
     has_location = False
     source = "https://covid19.ncdhhs.gov/dashboard/vaccinations"
--- a/can_tools/scrapers/official/SD/sd_vaccines.py
+++ b/can_tools/scrapers/official/SD/sd_vaccines.py
@@ -99,6 +99,34 @@ class SDVaccineCounty(MicrosoftBIDashboard):
                     }
                 }
             },
+            {
+                # TODO(sean): This remove booster shots from request. Currently treating boosters as a subset of
+                # "vaccinations completed" so they are not tracked.
+                "Condition": {
+                    "Not": {
+                        "Expression": {
+                            "In": {
+                                "Expressions": [
+                                    {
+                                        "Column": {
+                                            "Expression": {
+                                                "SourceRef": {"Source": "v"}
+                                            },
+                                            "Property": "Manufacturer - Dose # (spelled out)",
+                                        }
+                                    }
+                                ],
+                                "Values": [
+                                    [
+                                        {"Literal": {"Value": "'MOD3'"}},
+                                    ],
+                                    [{"Literal": {"Value": "'PFR3'"}}],
+                                ],
+                            }
+                        }
+                    }
+                }
+            },
         ]
 
         body["queries"] = [


can_tools/bootstrap_data/covid_demographics.csv
```diff
@@ -1,49 +1,58 @@
 age,race,ethnicity,sex
```
smcclure17, line 1: this adds a couple of rows for demographics in the DC scraper and sorts the rows

can_tools/scrapers/official/DE/de_vaccine.py
```diff
@@ -4,6 +4,7 @@
 from bs4 import BeautifulSoup
```
smcclure17, line 4: Delaware restructured their pages (moved all of the demographic charts onto one page) so this updates the scraper to match

can_tools/scrapers/official/NC/nc_vaccine.py
```diff
@@ -26,6 +26,12 @@ class NCVaccine(TableauDashboard):
         "AGG(Calc.Fully Vaccinated)-alias": variables.FULLY_VACCINATED_ALL,
     }
 
+    def fetch(self):
```
smcclure17, line 29: our native `TableauDashboard` class was throwing errors that I couldn't quickly track down. The `TableauScraper` module did not have the same problems so I switched the `fetch()` method to use it instead.

can_tools/bootstrap_data/covid_demographics.csv
mikelehen: SG. Part of me wonders if we could avoid having to always update this manually somehow, but as long as it's not too frequent, it's probably fine.

can_tools/scrapers/official/DE/de_vaccine.py
```diff
@@ -55,107 +53,98 @@ def normalize(self, data: List[pd.DataFrame]) -> pd.DataFrame:
...
                 # find the divs that contain the data tables
                 soup = BeautifulSoup(response.text, "lxml")
-                divs = soup.find_all("div", class_="c-table-with-chart")
+                divs = soup.find_all("div", class_="c-table-with-chart__wrapper")
 
                 for div in divs:
                     # find the div that contains the correct demographic data
-                    title = div.find("h2", text=re.compile(f"by {demographic}"))
+                    title = div.find(
+                        "h3", text=re.compile(f"% of Delawareans by {var} who {dose}")
```
mikelehen, line 94: And now I know what people from Delaware are called!

can_tools/scrapers/official/SD/sd_vaccines.py
```diff
@@ -99,6 +99,34 @@ def construct_body(self, resource_key, ds_id, model_id, report_id, counties):
                     }
                 }
             },
+            {
+                # TODO(sean): This remove booster shots from request. Currently treating boosters as a subset of
+                # "vaccinations completed" so they are not tracked.
```
mikelehen, line 104: Wow.  Out of curiosity is there a trick to "writing" these queries or do you do it by hand?

smcclure17: For the most part, I just copy the structure of the network calls made by the dashboard and tweak/add any columns or parameters that are needed. 

for example, for this where clause I looked for a query that included a "where not" clause, copied the syntax, and just changed the column and values to what we needed to exclude

can_tools/bootstrap_data/covid_demographics.csv
smcclure17: I think it's good to have as validation to ensure that the demographic pairs we collect are standardized. Like if a scraper started returning new data with a pair `all,all,all,non_hispanic` we would want to group that with `all,all,all,non-hispanic` instead of creating a new entry. 

It also does need to be updated pretty infrequently (once every month or two)

PR #361, opened 2021-11-01
replace CDCCovidDataTracker with official CDC testing dataset

smcclure17:
Creates a scraper for the new CDC testing dataset.

With this dataset, I think `CDCCovidDataTracker()` becomes obsolete. The new dataset does not track rolling average deaths, or rolling average total tests as the internal data tracker did, but this data was never used anywhere in the pipeline anyway. 

The data in this [historical dataset](https://data.cdc.gov/Public-Health-Surveillance/United-States-COVID-19-County-Level-of-Community-T/nra9-vzzn) lags behind the ["as originally posted"](https://data.cdc.gov/Public-Health-Surveillance/United-States-COVID-19-County-Level-of-Community-T/8396-v7yb) dataset, but includes history going back to the start of 2021 (as opposed to august 2021) and accounts for corrections, updates and other changes in the data over time (whereas the other only reports the initially reported values.)

With this update I think the only necessary change will be to [change the provider](https://github.com/covid-projections/covid-data-model/compare/smcclure17-update-test-source) of the CDC testing dataset source in covid-data-model. If this is checked in, I can create a test snapshot.

### DIFF
--- a/can_tools/scrapers/base.py
+++ b/can_tools/scrapers/base.py
@@ -609,25 +609,6 @@ class DatasetBase(ABC):
         "Internal _put method for dumping data using TempTable class"
         pass
 
-    def find_unknown_variable_id(self, engine: Engine, df: pd.DataFrame):
-        variables = pd.read_sql("select * from covid_variables", engine)
-        merged = df.merge(variables, on=["category", "measurement", "unit"], how="left")
-        bad = merged["id"].isna()
-        return df.loc[bad, :]
-
-    def find_unknown_location_id(self, engine: Engine, df: pd.DataFrame):
-        locs = pd.read_sql("select * from locations", engine)
-        good_rows = df.location_name.isin(
-            locs.loc[locs.state_fips == self.state_fips, :].name
-        )
-        return df.loc[~good_rows, :]
-
-    def find_unknown_demographic_id(self, engine: Engine, df: pd.DataFrame):
-        dems = pd.read_sql("select * from covid_demographics", engine)
-        merged = df.merge(dems, on=["sex", "age", "race", "ethnicity"], how="left")
-        bad = list(merged["id"].isna())
-        return df.loc[bad, :]
-
     def fetch_normalize(self):
         "Call `self.normalize(self.fetch())`"
         data = self.fetch()
--- a/can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
+++ b/can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
@@ -15,6 +15,9 @@ class CDCHistoricalTestingDataset(FederalDashboard):
     location_type = "county"
     source = "https://data.cdc.gov/Public-Health-Surveillance/United-States-COVID-19-County-Level-of-Community-T/nra9-vzzn/data"
     source_name = "Centers for Disease Control and Prevention"
+    # We also collect CDC testing data via the CDCCovidDataTracker class. 
+    # In order to not overwrite/mix the data sources we use the cdc2 provider instead of cdc.
+    # 11/1/21: This is the offical CDC testing dataset and the one that is used by the pipeline downstream.
     provider = "cdc2"
 
     # SODA API stream has different variable names than the CSV and online dataset -- but the columns appear to be the same.
--- a/can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
+++ b/can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
@@ -1,0 +1,59 @@
+from typing import Dict
+
+import pandas as pd
+
+from can_tools.scrapers.base import CMU
+from can_tools.scrapers.official.base import FederalDashboard
+
+# Set the maximum number of records to ten million to ensure we collect all rows in the dataset.
+# Currently the size of the dataset is 2 million records, so this may need to be updated (increased)
+# in the future
+SODA_API_RESPONSE_LIMIT = 10000000
+
+class CDCHistoricalTestingDataset(FederalDashboard):
+    has_location = True
+    location_type = "county"
+    source = "https://data.cdc.gov/Public-Health-Surveillance/United-States-COVID-19-County-Level-of-Community-T/nra9-vzzn/data"
+    source_name = "Centers for Disease Control and Prevention"
+    provider = "cdc2"
+
+    # SODA API stream has different variable names than the CSV and online dataset -- but the columns appear to be the same.
+    variables = {
+        "cases_per_100k_7_day_count": CMU(
+            category="cases", measurement="rolling_average_7_day", unit="people"
+        ),
+        "percent_test_results_reported": CMU(
+            category="pcr_tests_positive",
+            measurement="rolling_average_7_day",
+            unit="percentage",
+        ),
+    }
+
+    def fetch(self) -> pd.DataFrame:
+        # select only the columns we care about in order to speed up query
+        data = pd.read_json(
+            f"https://data.cdc.gov/resource/nra9-vzzn.json?$limit={SODA_API_RESPONSE_LIMIT}"
+            "&$select=fips_code,date,cases_per_100k_7_day_count,percent_test_results_reported"
+        )
+        if len(data) == SODA_API_RESPONSE_LIMIT:
+            raise ValueError(
+                f"Size of the dataset has reached the currently set API response limit of {SODA_API_RESPONSE_LIMIT}. "
+                "Please increase the limit by modifying SODA_API_RESPONSE_LIMIT."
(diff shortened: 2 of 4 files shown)


services/prefect/flows/generated_flows.py
```diff
@@ -97,46 +94,6 @@ def create_flow_for_scraper(ix: int, cls: Type[DatasetBase], schedule=True):
     return flow
 
 
-def create_cdc_single_state_flow():
```
smcclure17, line 100: Removing the CDCCovidDataTracker removes the need for this flow. This would create a task to run `CDCCovidDataTracker` for each state.

tests/test_datasets.py
```diff
@@ -57,10 +54,7 @@ def test_datasets(cls):
     execution_date = (pd.Timestamp.today() - pd.Timedelta("1 days")).strftime(
         "%Y-%m-%d"
     )
-    if cls == CDCCovidDataTracker:
```
smcclure17, line 60: See above comment on removing this scraper

can_tools/scrapers/base.py
```diff
@@ -617,6 +617,9 @@ def find_unknown_variable_id(self, engine: Engine, df: pd.DataFrame):
 
     def find_unknown_location_id(self, engine: Engine, df: pd.DataFrame):
         locs = pd.read_sql("select * from locations", engine)
+        if "location" in df.columns:
```
smcclure17, line 620: Not related to the scraper, but this function helps find any missing/unknown locations in the scraper output. Previously it only worked on scrapers that track locations by county name, so this adds the ability to find unknown fips codes, too.

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
```diff
@@ -0,0 +1,60 @@
...
+# in the future
+SODA_API_RESPONSE_LIMIT = 10000000
+
+
+class CDCHistoricalTestingDataset(FederalDashboard):
+    has_location = True
+    location_type = "county"
+    source = "https://data.cdc.gov/Public-Health-Surveillance/United-States-COVID-19-County-Level-of-Community-T/nra9-vzzn/data"
+    source_name = "Centers for Disease Control and Prevention"
+    provider = "cdc2"
```
smcclure17, line 19: I used cdc2 here (as with the historical vaccine data) to not mix up this data with the previously collected CDC Covid Tracker data. 

I think that it would be okay to have this data override the old CDC data, but I'm concerned that since the new dataset and the old CDC Tracker will have different `source`/`source_name` columns, I think we will end up with rows that are entirely duplicated except for the `source` and `source_name` columns, and I'm not sure how the pipeline would handle that

smcclure17: If we _were_ to use the `cdc` provider here (and did not run into issues with duplicate entries) then I don't think we would need to change the downstream pipeline at all

can_tools/scrapers/base.py
```diff
@@ -617,6 +617,9 @@ def find_unknown_variable_id(self, engine: Engine, df: pd.DataFrame):
...
+            good_rows = df.location.isin(locs.location)
+            return df.loc[~good_rows, :]
```
mikelehen, line 622: Hrm.  The below (`location_name`) version of this code seems to filter to locations in `self.state_fips` so I'm wondering if we should do the same for this new `location` version?

That said I actually don't see `state_fips` defined / used on this class so I'm a bit confused. Not sure where it comes from. :/

can_tools/scrapers/official/federal/CDC/cdc_coviddatatracker.py
```diff
@@ -1,108 +0,0 @@
...
-        "new_deaths_7_day_rolling_average": CMU(
-            category="deaths", measurement="rolling_average_7_day", unit="people"
-        ),
-        "percent_new_test_results_reported_positive_7_day_rolling_average": CMU(
-            category="pcr_tests_positive",
-            measurement="rolling_average_7_day",
-            unit="percentage",
-        ),
-        "new_test_results_reported_7_day_rolling_average": CMU(
-            category="pcr_tests_total",
```
mikelehen, line 33: I just sent email to the CMU folks, but I think they are relying on this field.  So we may need to keep this scraper. 🙁

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
```diff
@@ -0,0 +1,60 @@
...
+        # select only the columns we care about in order to speed up query
+        data = pd.read_json(
+            f"https://data.cdc.gov/resource/nra9-vzzn.json?$limit={SODA_API_RESPONSE_LIMIT}"
+            "&$select=fips_code,date,cases_per_100k_7_day_count,percent_test_results_reported"
+        )
+        if len(data) == SODA_API_RESPONSE_LIMIT:
+            raise ValueError(
+                f"Size of the dataset has reached the currently set API response limit of {SODA_API_RESPONSE_LIMIT}. "
+                "Please increase the limit by modifying SODA_API_RESPONSE_LIMIT."
+            )
```
mikelehen, line 43: Nice, thanks for putting this sanity-check in!

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
mikelehen: Hrm.  What's the primary key for rows in the database?  I would assume that if we used `cdc` then we'd just overwrite existing rows.  I guess the main risk would be if the new dataset had fewer historical datapoints than the old data.  We'd have a mix of sources in the resulting data, depending on the date?  In that case, the website might end up linking to the wrong place, I guess.  And the bigger issue would be that we might have jumps in the data if the two datasets aren't compatible.  To be safe we would delete the old data.

That said, per my other comment I think we may need to keep the old scraper and so making this be `cdc2` makes sense.  But let's add a comment to try to explain a bit of context.

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
```diff
@@ -0,0 +1,60 @@
...
+        return data
+
+    def normalize(self, data: pd.DataFrame) -> pd.DataFrame:
+        return (
+            # remove suppressed values as we have no way to handle them
+            data.loc[~data.eq("suppressed").any(1)]
```
mikelehen, line 49: Just for my own curiosity (I'm not too worried about it), does this mean that if either `cases_per_100k_7_day_count` or `percent_test_results_reported` is suppressed, then we will end up dropping the whole row (both variables)?

services/prefect/flows/generated_flows.py
mikelehen: Ah, yeah...  It would be nice to delete this.  But I don't think we should, at least until we hear back from CMU.  😦

mikelehen (approved):
Thanks for pulling this together so fast!  Unfortunately I think CMU is relying on the `pcr_tests_total` column from the old dataset so I think we need to keep it around for now.  I did email them to double-check though.  Maybe they don't need it and we can delete it.

services/prefect/flows/generated_flows.py
smcclure17: Very good call, I had forgotten about that group. I'll re-add this!

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
smcclure17: ah, this would and it shouldn't -- good catch I'll fix this

can_tools/scrapers/base.py
smcclure17: Yeah, this should probably either be a method of the [`StateDashboard`](https://github.com/covid-projections/can-scrapers/blob/main/can_tools/scrapers/official/base.py#L35) class (which has the `state_fips` attribute) or a stand-alone utils function. 

The `location_name` version specifies a state b/c states can have similarly named counties (eg, there are a million Washington counties). This limits the search to within the bounds of the state of the scraper (I think it was assumed that this would only be used on single state scrapers, but I'm not sure.)

Since the `location` version searches for FIPS codes, which are unique, specifying the state shouldn't be needed.

mikelehen: Ahhh...  Yeah, that makes total sense. Thanks!

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
smcclure17: good question on the primary key 😅 I'm not actually sure -- I'll take a look

PR #363, opened 2021-11-03
scraper fixes for GA, SD, DE, CDC, NY, AZ Maricopa

smcclure17:


can_tools/scrapers/__init__.py
```diff
@@ -193,7 +193,6 @@
 )
 from can_tools.scrapers.official.WY.WYVaccines import (
     WYCountyVaccinations,
-    WYStateVaccinations,
```
smcclure17, line 196: this scraper keeps breaking, and since we don't currently, and have never used (afaik), state-sourced state-level data, I want to just remove this scraper

can_tools/scrapers/official/DE/de_vaccine.py
```diff
@@ -144,6 +144,7 @@ def normalize(self, data: Dict[str, requests.models.Response]) -> pd.DataFrame:
                 vintage=self._retrieve_vintage(),
                 location_name=lambda row: row["location_name"].str.replace("-", " "),
                 age=lambda row: row["age"].str.replace("+", "_plus"),
+                dt=lambda row: row["dt"].dt.date
```
smcclure17, line 147: The date column was throwing an error about datatypes (specifically `datetime64[ns, UTC]`), so this forces it into a workable format

can_tools/scrapers/official/GA/ga_vaccines.py
```diff
@@ -18,39 +18,29 @@ class GeorgiaCountyVaccine(StateDashboard):
     fetch_url = "https://georgiadph.maps.arcgis.com/sharing/rest/content/items/e7378d64d3fa4bc2a67b2ea40e4748b0/data"
 
     variables = {
-        "CUMPERSONCVAX": variables.FULLY_VACCINATED_ALL,
-        "CUMPERSONVAX": variables.INITIATING_VACCINATIONS_ALL,
+        "PERSONCVAX": variables.FULLY_VACCINATED_ALL,
```
smcclure17, line 21: The data in the timeseries sheets, which is what we were using previously, looks wrong. For many counties, fully-vaccinated was greater than vaccinations-initiated. 

This somewhat draws the other data in their export into question, but the "county summary" sheet (that does not have historical data) does not seem to have the same issues.

As such, this pulls from the summary sheet, which means we lose timeseries data going forward. If the GA CDC data weren't so bad, I would recommend switching over

can_tools/scrapers/official/OH/oh_vaccine_demographics.py
```diff
@@ -54,9 +54,11 @@ def _extract_data(self, shot_type: str) -> pd.DataFrame:
         for county in counties:
             print("working on county: ", county)
             wb = ws.select("county", county)
-            df = wb.getWorksheet("New Demographics").data.assign(location_name=county)
+            try:
+                df = wb.getWorksheet("New Demographics").data.assign(location_name=county)
+            except TypeError:
+                print(f"Could not fetch data for {county}, skipping...")
```
smcclure17, line 60: Sometimes, one or two counties are not included in the data tables (I'm presuming b/c they did not report data, but not sure.) 

Currently, if any counties are missing the scraper breaks. This is great as it lets us know when any of the counties are missing, but it also means that if we're missing data for a single county then we don't get data for any counties :/ 

This lets the scraper continue without the missing counties, which I'm a little apprehensive about because we won't be alerted if more and more counties don't have data. But, I'm not sure of a great alternative. We could potentially set up a logger for the Prefect tasks but I'm not sure if that's worth the effort.

can_tools/scrapers/official/WY/WYVaccines.py
```diff
@@ -11,87 +11,7 @@
 
 from can_tools.scrapers.base import CMU
 from can_tools.scrapers import variables
-from can_tools.scrapers.official.base import GoogleDataStudioDashboard, StateDashboard
```
smcclure17, line 14: see above about removing this scraper

can_tools/scrapers/official/SD/sd_vaccines.py
```diff
@@ -99,34 +99,6 @@ def construct_body(self, resource_key, ds_id, model_id, report_id, counties):
                     }
                 }
             },
-            {
```
smcclure17, line 102: So, we actually need these booster shots to calculate the other doses correctly. 

SD defines each bucket as the number of people who currently have the bucket-specific number of doses. So someone in the "1-dose" Moderna bucket only has 1 dose, and when they receive their second dose they are removed from this bucket and added to the "2-dose" bucket. 

The same goes for boosters, so we need to add the booster shots to the existing dose buckets to get accurate fully vaccinated and fully initiated counts.

smcclure17: Also, FWIW the dashboard looks like it has some problems. It's displaying negative values for the manufacturer-specific data. The underlying data (that we access) seems to be okay, but it is a little concerning.

image

can_tools/scrapers/official/GA/ga_vaccines.py
mikelehen: Ugh. :/ Thanks for digging into it and explaining.

can_tools/scrapers/official/OH/oh_vaccine_demographics.py
mikelehen: Yeah, this seems fine for now.  This is kind of a systemic problem (the compare tool doesn't highlight when locations stop updating until they disappear completely :-/).

can_tools/scrapers/official/SD/sd_vaccines.py
mikelehen: Oh, wow.  That's tricky / annoying.  Nice work detangling this.

can_tools/scrapers/official/SD/sd_vaccines.py
```diff
@@ -221,15 +191,30 @@ def normalize(self, resjson):
...
+            df["moderna_1_dose"] + df["pfizer_1_dose"] + df["total_vaccine_completed"]
+        )
+
         df["total_vaccine_doses_administered"] = (
             df["janssen_series"]
             + df["moderna_1_dose"]
             + df["pfizer_1_dose"]
             + 2 * (df["moderna_complete"] + df["pfizer_complete"])
+            + 3 * (df["moderna_booster"] + df["pfizer_booster"])
+            + 2 * (df["janssen_booster"])
```
mikelehen, line 217: FWIW- This is going to have some error in it because you can mix/match vaccines and boosters.

E.g. 1 janssen + 1 moderna booster (we'll count as 3 but should be 2) or 2 moderna + 1 janssen booster (we'll count as 2 but should be 3).

Might at least be worth a comment.  Or we could just drop `total_vaccine_doses_administered` since our pipeline doesn't need it.

mikelehen (approved):
Looks good.  Thanks for all the fixing! 😬

can_tools/scrapers/official/SD/sd_vaccines.py
smcclure17: ah, good flag -- I'll just remove it!

PR #364, opened 2021-11-10
add scraper for OR race, sex, and age

smcclure17:
data for 1+ dose only, sourced from:
https://public.tableau.com/app/profile/oregon.health.authority.covid.19/viz/OregonCOVID-19VaccinationTrends/OregonStatewideVaccinationTrends

I compared Multnomah County to the output of the scraper to verify the results

### DIFF
--- a/can_tools/utils.py
+++ b/can_tools/utils.py
@@ -2,6 +2,7 @@ from typing import Optional
 import pandas as pd
 import sqlalchemy as sa
 from sqlalchemy.engine.base import Engine
+from can_tools.models import create_dev_engine
 
 
 def determine_location_column(df: pd.DataFrame) -> str:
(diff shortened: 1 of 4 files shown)


can_tools/utils.py
```diff
@@ -2,6 +2,7 @@
 import pandas as pd
```
smcclure17, line 2: I forgot this was in here -- 

this restructures some of the helper functions

mikelehen (approved):
Looks good! I'm going to trust you to sanity-check these (which it sounds like you already did!) but if you'd like me to do any additional testing, let me know and I can.

smcclure17:
sounds good -- for these tableau dashboards we extract the tables directly (well, as directly as they let us) from the dashboard. So, the numbers _should_ always match the dashboard, and the main question is whether the definitions line up

PR #365, opened 2021-11-10
add Michigan demographics

smcclure17:
data from https://www.michigan.gov/coronavirus/0,9753,7-406-98178_103214_103272-547150--,00.html

for intersections of age and race

EDIT:

I'm realizing, we do we know if the covid-data-model pipeline will aggregate these intersections into individual, respective demographic buckets? The API provides `vaccinationsInitiatedDemographics`, `age`, and `race` categories separately.

I worry that the pipeline might not be able to use this format--Should I aggregate this data to get individual `race` and `age` buckets? (e.g. `12-15,all,all,all` instead of `12-15,ai_an,all,all`)


mikelehen:
@BrettBoval can hopefully confirm, but I think Surgo may be pulling data directly from the parquet file rather than going through our API so I am not sure we need to worry about covid-data-model limitations.

smcclure17:
cool, I hope they do haha 😆 

I reached out to Nick, I'm going to leave this PR here as is until I hear back

mikelehen (approved):
Looks good!

PR #366, opened 2021-11-11
add SC demographic scraper

smcclure17:
this also collects intersections of demographics (race,age, and sex) so we'll need to wait until I hear back from Nick to land this


can_tools/scrapers/official/SC/sc_demographics.py
```diff
@@ -0,0 +1,83 @@
...
+
+        engine = engine.getWorksheet("Vaccine Map By SC residents PEOPLE")
+        filters = engine.getFilters()
+        counties = [
+            t["values"] for t in filters if t["column"] == "Recipient County for maps"
+        ][0]
+
+        data = []
+        for county in counties:
+            print("parsing", county)
```
mikelehen, line 42: Same question here but now that I've seen it in 2 scrapers I'm guessing it's intentional. 😁

PR #367, opened 2021-11-11
add race scrapers for AZ and Maricopa County, AZ

smcclure17:


can_tools/scrapers/official/AZ/az_demographics.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # extract data from Tableau Dashboard
+        counties = [
+            t["values"]
+            for t in worksheet.getSelectableItems()
+            if t["column"] == "Admin Address County W/ State Pod"
+        ][0]
+        raw_data = []
+        for county in counties:
+            print("parsing", county)
```
mikelehen, line 38: Was this just for debugging? Or do we want it checked-in (no preference on my end; just double-checking)?

can_tools/scrapers/official/AZ/az_demographics.py
```diff
@@ -0,0 +1,83 @@
...
+            workbook = worksheet.select("Admin Address County W/ State Pod", county)
+            raw_data.append(
+                workbook.getWorksheet("Race (# people)").data.assign(
+                    location_name=county
+                )
+            )
+        data = pd.concat(raw_data)
```
mikelehen, line 45: nit: I assume concat isn't actually needed here?

can_tools/scrapers/official/AZ/az_demographics.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+            .pipe(
+                self.extract_CMU, cmu=self.variables, skip_columns=["race", "ethnicity"]
+            )
+        )
+
+        # All the race demographics specify that the entries are non-hispanic
+        # except for 'Hispanic or Latino' value.
+        # So, by default set ethnicity as non-hispanic (done above), and set to hispanic where race == Hispanic or Latino (done below)
+        # When race is unknown, ethnicity is also unknown, modify columns to reflect this.
```
mikelehen, line 80: nit: I was confused by setting ethnicity to non-hispanic above until I got to here.  Maybe move or split this comment so that somebody reading this code beginning to end will have an easier time understanding what's going on and knowing it's intentional, not a bug?

mikelehen (approved):
LGTM with minor optional nits.

can_tools/scrapers/official/AZ/az_demographics.py
smcclure17: I think its needed as `raw_data` is a list of dataframes that then get combined into one here. We could probably make `raw_data` into a dataframe initially, and have the append in line 40 append to the dataframe directly (I'm not sure how that would compare performance wise). But, I think I'm going to leave it as-is for now if that's okay

can_tools/scrapers/official/AZ/az_demographics.py
smcclure17: good call, will do

can_tools/scrapers/official/AZ/az_demographics.py
smcclure17: mostly just for debugging to make sure the scraper is working -- I'll remove

can_tools/scrapers/official/AZ/az_demographics.py
mikelehen: Ah, whoops! Yeah, I assumed wrong (I missed that raw_data was a list). My bad, thanks for explaining.  Carry on!

PR #368, opened 2021-11-11
Add NH vaccinations by race

smcclure17:


can_tools/scrapers/official/NH/nh_demographics.py
```diff
@@ -0,0 +1,128 @@
...
+        # fetch NUM_PROCESSES counties at a time
+        pool = multiprocessing.Pool(processes=NUM_PROCESSES)
+        data = []
+
+        # Setting multiple parameter values on the same parameter type
+        # on a single TableauScraper instance removes the
+        # other selections (in this case, the county selection) from the instance, and there doesn't seem
+        # to be a good way to fix/circumvent this. So, we use two TableauScraper
+        # instances to fetch each respective dose type.
+        # This is an issue we've had in the past with the OH and WI demographic scrapers.
```
mikelehen, line 52: Thanks for the comment!

can_tools/scrapers/official/NH/nh_demographics.py
```diff
@@ -0,0 +1,128 @@
...
+        raw_data.append(
+            workbook.getWorksheet("Count: Bar Chart").data.assign(location_name=county)
+        )
+
+        data = pd.concat(raw_data)
+        return data
+
+    def normalize(self, data) -> pd.DataFrame:
+        demo_replace = {
+            "Hispanic/Latino": "hispanic",
```
mikelehen, line 83: Per slack, maybe move this to the ethnicity column?

can_tools/scrapers/official/NH/nh_demographics.py
```diff
@@ -0,0 +1,128 @@
...
+    location_type = "county"
+
+    state_fips = us.states.lookup("New Hampshire").fips
+
+    variables = {
+        "Residents with At Least 1 Dose:": variables.INITIATING_VACCINATIONS_ALL,
+        "Residents Fully Vaccinated:": variables.FULLY_VACCINATED_ALL,
+    }
+
+    # hard coding to make parallel fetching easier
```
mikelehen, line 26: mostly for my curiosity, why does this make it easier?  What would break if you just did `counties = self._retrieve_counties()` in the fetch() method?

can_tools/scrapers/official/NH/nh_demographics.py
```diff
@@ -0,0 +1,128 @@
...
+        data = data.loc[:, cols_to_keep]
+        data["location_name"] = data["location_name"].replace(
+            {
+                "Rest of Hillsborough": "Hillsborough",
+                "Nashua": "Hillsborough",
+                "Manchester": "Hillsborough",
+            }
+        )
+        group = [col for col in cols_to_keep if col != "value"]
+        return data.groupby(group).sum().reset_index()
```
mikelehen: Sometime when we're moving at a more leisurely pace it might be nice to create a helper for this location merging thing...

mikelehen (approved):
The Tableau part is mostly gobbledygook to me, but the rest looks reasonable! 😁

can_tools/scrapers/official/NH/nh_demographics.py
smcclure17: good call, just changed that 🙏

can_tools/scrapers/official/NH/nh_demographics.py
smcclure17: In this case there are some locations that don't exactly match the column names in the dashboard (because of how they split up Hillsborough into "rest of Hillsborough", "Nashua", and "Manchester"). Using self._retrieve_counties(), we wouldn't collect those and would be missing those locations. 

We could call `self._retrieve_counties()` and edit/append to the output to match these different locations, but I figured it was less work to just hard code them. (These hard coded values come directly from the dashboard so they match 1 to 1 with the TableauScraper selects/queries)

can_tools/scrapers/official/NH/nh_demographics.py
smcclure17: yeah, I agree

PR #370, opened 2021-11-12
Add ILVaccineAge scraper.

mikelehen:
This was quick to do so I just went ahead.

Ideally I'd probably share code between the two, but it's not that much code so probably fine?  I can clean it up if you want.


mikelehen:
Actually, the age groups are just 5-11, 12-17, 18-64 and 65+ which is basically the same as we get from CDC.  So I'll just close this.

PR #371, opened 2021-11-12
Surgo nc demos

mikelehen:


can_tools/scrapers/official/NC/nc_demographics.py
```diff
@@ -0,0 +1,35 @@
...
+        engine = TableauScraper()
+        engine.loads(
+            "https://public.tableau.com/views/NCDHHS_COVID-19_Dashboard_Vaccinations/VaccinationDashboard"
+        )
+        engine = engine.getWorkbook()
+        sp = engine.goToStoryPoint("2")
+        wb = sp.setParameter("County", "Wake County")
+
+        # i feel like with the set paramater that this should be returning data in the Race_county... tables
+        # but nothing :/
```
mikelehen, line 34: @smcclure17 Not sure if it helps but while I don't see anything under "Race_county", I do see data under "Race_Percent_county"...  But this is % as in "The percent of the black population that is fully vaccinated" which I'm not sure we can / want to represent?

Anyway, not sure if this is helpful, but thought I'd mention it in case it is.

smcclure17: Yeah, I'm not sure that's useful on its own, (without population data that we can multiply by), but it is promising that there's some sort of county-level data there. Seems like if that data is here then the rest must be somewhere, but not sure where :/

I think I'm going to table this for now if that's ok, and try to come back to it whenever I find some free time.

Thanks for taking a look into it! 😃

mikelehen: FYI- I poked at this a tiny bit more and I think it *might* have raw counts too, but I'm not 100% sure if that's what these numbers are or not.  We'd need to sanity-check.

image

In any case, I'm 100% okay with tabling this.

PR #372, opened 2021-11-15
Scraper fixes and changes to requirements.txt

smcclure17:
a batch of scraper fixes that I fell behind on last week.

### DIFF
--- a/can_tools/bootstrap_data/covid_demographics.csv
+++ b/can_tools/bootstrap_data/covid_demographics.csv
@@ -388,3 +388,9 @@ all,asian_or_pacific_islander,non-hispanic,all
 55-64,white,all,female
 65_plus,white,all,male
 65_plus,white,all,female
+5-11,ai_an,all,all
+12-19,all,all,all
+5-11,all,all,female
+5-11,all,all,male
+5-11,all,hispanic,all
+5-11,all,non-hispanic,all
--- a/can_tools/scrapers/official/DE/de_vaccine.py
+++ b/can_tools/scrapers/official/DE/de_vaccine.py
@@ -141,7 +141,7 @@ class DelawareVaccineDemographics(DelawareCountyVaccine):
                 timezone="US/Eastern",
             )
             .assign(
-                vintage=self._retrieve_vintage(),
+                vintage=self._retrieve_vintage().tz_localize(None),
                 location_name=lambda row: row["location_name"].str.replace("-", " "),
                 age=lambda row: row["age"].str.replace("+", "_plus"),
                 dt=lambda row: row["dt"].dt.date,
--- a/can_tools/scrapers/official/IL/il_vaccine.py
+++ b/can_tools/scrapers/official/IL/il_vaccine.py
@@ -28,7 +28,7 @@ class IllinoisVaccineCounty(StateDashboard):
         return (
             # rename the Chicago rows such that they are summed together with the Cook County entries,
             # effectively merging the two locations into one (as CHI is within Cook County)
-            data.replace({"Chicago": "Cook"})
+            data.replace({"Chicago": "Cook", "Chicago ": "Cook"})
             .groupby(["CountyName", "Report_Date"])
             .sum()
             .reset_index()
--- a/can_tools/scrapers/official/ME/me_vaccines.py
+++ b/can_tools/scrapers/official/ME/me_vaccines.py
@@ -416,11 +416,7 @@ class MaineRaceVaccines(MicrosoftBIDashboard):
         # Build post headers
         headers = self.construct_headers(resource_key)
 
-        counties = list(
-            pd.read_csv(
-                os.path.dirname(__file__) + "/../../../bootstrap_data/locations.csv"
-            ).query("state == @self.state_fips and name != 'Maine'")["name"]
-        )
+        counties = self._retrieve_counties()
 
         jsons = []
         """
@@ -460,13 +456,19 @@ class MaineRaceVaccines(MicrosoftBIDashboard):
 
         # combine into dataframe
         df = pd.DataFrame.from_records(data_rows)
-        # they report j&j doses in either the initiated or completed column--the other is empty (NA)
-        df["jj_complete"] = df["jj_complete"].fillna(0)
-        df["jj_init"] = df["jj_init"].fillna(0)
+
+        # J&J doses are reported in either the initiated or completed column--the other is empty (NaN)
+        # NOTE(sean): 11/14/21, the 5-11 age group does not have J&J data (it's not authorized for those under 18),
+        # but the J&J columns need to exist for the rest of the method to work.
+        # This just adds an empty column for if the J&J columns do not exist.
+        if "jj_complete" not in df.columns:
+            df["jj_complete"] = 0
+        if "jj_init" not in df.columns:
+            df["jj_complete"] = 0
 
         # format, calculate total_vacccine_initiated + map CMU
         out = (
-            df.dropna()  # counties that don't have age 0-11 return NA for value
+            df.fillna(0)
             .assign(
                 initiated_total=lambda x: x["first_dose_total"]
                 + x["jj_complete"]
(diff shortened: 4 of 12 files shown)


can_tools/bootstrap_data/covid_demographics.csv
```diff
@@ -387,4 +387,10 @@ all,asian_or_pacific_islander,non-hispanic,all
 55-64,white,all,male
 55-64,white,all,female
 65_plus,white,all,male
-65_plus,white,all,female
\ No newline at end of file
+65_plus,white,all,female
+5-11,ai_an,all,all
```
smcclure17, line 391: Not sure if we want to zero-pad these values or not, but most dashboards seem to return values without padding (e.g "5-11" instead of "05-11") so I stuck with that. I think it is important that we keep this standardized either way

can_tools/scrapers/official/DE/de_vaccine.py
```diff
@@ -141,7 +141,7 @@ def normalize(self, data: Dict[str, requests.models.Response]) -> pd.DataFrame:
                 timezone="US/Eastern",
             )
             .assign(
-                vintage=self._retrieve_vintage(),
```
smcclure17, line 144: not 100% sure why this is needed, but the datatype before was `datetime64[ns, UTC]` which made the scraper fail on insert. This just removes the UTC

can_tools/scrapers/official/IL/il_vaccine.py
```diff
@@ -28,7 +28,7 @@ def normalize(self, data: pd.DataFrame) -> pd.DataFrame:
         return (
             # rename the Chicago rows such that they are summed together with the Cook County entries,
             # effectively merging the two locations into one (as CHI is within Cook County)
-            data.replace({"Chicago": "Cook"})
+            data.replace({"Chicago": "Cook", "Chicago ": "Cook"})
```
smcclure17, line 31: annoying space 🤕

can_tools/scrapers/official/ME/me_vaccines.py
```diff
@@ -460,13 +456,19 @@ def normalize(self, resjson: dict) -> pd.DataFrame:
...
+        # This just adds an empty column for if the J&J columns do not exist.
+        if "jj_complete" not in df.columns:
+            df["jj_complete"] = 0
+        if "jj_init" not in df.columns:
+            df["jj_complete"] = 0
 
         # format, calculate total_vacccine_initiated + map CMU
         out = (
-            df.dropna()  # counties that don't have age 0-11 return NA for value
+            df.fillna(0)
```
smcclure17, line 471: I don't remember why we were dropping values NA values here, but I think it makes more sense to just mark them as 0s (especially since we were already doing that with the J&J doses above)

can_tools/scrapers/official/TX/texas_vaccine.py
```diff
@@ -142,11 +145,14 @@ class TXVaccineCountyAge(TexasVaccineParent):
...
-            "16-49 years": "16-49",
-            "50-64 years": "50-64",
-            "65-79 years": "65-79",
-            "80+ years": "80_plus",
+            # Excel automatically parses some of the string entries as dates, and
+            # read excel does not seem to have a way to fix this.
+            # This is not a good fix, but it's the only way I could find to convert the
+            # date values back into strings.S
+            # See https://github.com/pandas-dev/pandas/issues/39898 for explanation
+            "2021-05-11 00:00:00": "5-11",
```
smcclure17, line 153: Excel was doing some fun excel stuff here with assuming the datatypes--it interpreted `12-15` and `5-11` as dates. I did a bit of digging but couldn't find a real fix, so I opted for this kinda' gross solution. 

This issue is related, but didn't provide a good solution: https://github.com/pandas-dev/pandas/issues/39898

can_tools/scrapers/official/VT/vt_vaccinations.py
```diff
@@ -1,15 +1,11 @@
+from typing import Dict
```
smcclure17, line 1: Looks like Vermont restructured the underlying data for their dashboard. Unfortunately, I couldn't find raw cumulative counts anywhere in the data, but they did provide %s and population values, so I use that to calculate cumulative counts.

can_tools/utils.py
```diff
@@ -70,11 +70,19 @@ def find_unknown_location_id(df: pd.DataFrame, state_fips: int, engine: Engine =
...
+def find_unknown_demographic_id(
+    df: pd.DataFrame, engine: Engine = None, csv_rows=False
+):
     """Find any demographic pairs in the specified dataframe that do not match an entry in the covid_demographics file"""
     if not engine:
         engine = create_dev_engine()[0]
     dems = pd.read_sql("select * from covid_demographics", engine)
     merged = df.merge(dems, on=["sex", "age", "race", "ethnicity"], how="left")
     bad = list(merged["id"].isna())
-    return df.loc[bad, :]
```
smcclure17, line 80: let's us output the missing demographic pairs as CSV strings like `5-11,all,all,all\n12-19,all,all,all`.

It's still a bit of work as you have to "find and replace" the \n's into actual new lines, but it does make it easier to add the new demographics to the file.

It would be possible (I think) to have this write directly to the `covid_demographics.csv` file, appending any new demographic pairs, but I think having the manual validation is a good thing.

requirements.txt
```diff
@@ -12,7 +12,7 @@ sqlalchemy==1.3.20
 sqlalchemy_utils==0.36.8
 camelot-py[cv]
 us==2.0.2
-TableauScraper==0.1.2
+TableauScraper==0.1.29
```
smcclure17, line 15: I'm not sure what changes were made, but the NH and SC scrapers need this update to succeed

can_tools/bootstrap_data/covid_demographics.csv
mikelehen: 👍  no zero-padding sounds good to me.

can_tools/scrapers/official/DE/de_vaccine.py
mikelehen: Hrm.  That's really strange.  From a quick search, we have ~50 scrapers that have code like this and seem to be working fine.  Any idea at all why this one is different? It makes me a bit nervous.

can_tools/scrapers/official/TX/texas_vaccine.py
mikelehen: Heh, yikes.  Thanks for the comment.

can_tools/scrapers/official/VT/vt_vaccinations.py
```diff
@@ -69,21 +57,23 @@ def construct_body(self, resource_key, ds_id, model_id, report_id, county):
...
+                                            ("c", "County", "location_name")
                                         ],
-                                        [],
+                                        [("b", "Count", 0, "count")],
                                         [
                                             # Measures
-                                            ("_1", "People vaccinated", "init"),
-                                            ("i", "People completed 1", "completed"),
+                                            ("m", "Map % vaccinated", "init"),
+                                            # ("i", "People completed 1", "completed"),
```
mikelehen, line 76: Remove this or leave a comment for why it's here but commented out?

can_tools/utils.py
mikelehen: Nice!  Yeah, I agree it should be manual.  I was wondering though how hard it would be to make the tests catch the sql constraint blah blah errors and then automatically run something like this and say, "These rows were not found in covid_demographics.csv. Verify they're expected and then add them: ..."  Not urgent or anything though.

requirements.txt
mikelehen: Yeah. I noticed this when I was poking at the NC one. 🤷

mikelehen (approved):
Nice! Thanks as always for dealing with all these!

can_tools/scrapers/official/DE/de_vaccine.py
smcclure17: Yeah I'm with you--I'm really not sure. It calls the `_retrieve_vintage()` method directly, so it doesn't seem like there's any reason for this to be different than the others :/ 

I can put in some more effort to try and see what's going on

can_tools/utils.py
smcclure17: Yeah I like the sound of that--I'll open a PR after this to add that

PR #374, opened 2021-11-15
Fix and update NH, DE, NY, MI scrapers

smcclure17:


can_tools/scrapers/official/DE/de_vaccine.py
```diff
@@ -146,6 +145,9 @@ def normalize(self, data: Dict[str, requests.models.Response]) -> pd.DataFrame:
                 age=lambda row: row["age"].str.replace("+", "_plus"),
                 dt=lambda row: row["dt"].dt.date,
             )
-            .replace(["patient_declined_to_disclose", "data_not_reported"], None)
```
smcclure17, line 149: this was causing some duplicate rows, so I replaced it with the code below and verified that the data was correct

can_tools/scrapers/official/MI/mi_vaccine_demographics.py
```diff
@@ -63,9 +63,19 @@ def normalize(self, data: pd.DataFrame) -> pd.DataFrame:
...
         # combine detroit and wayne county into one location
         data["location_name"] = data["location_name"].replace("Detroit", "Wayne")
-        data = (
-            data.groupby(by=[col for col in data.columns if col != "value"])
-            .sum()
-            .reset_index()
-        )
-        return data
+
+        # sum over age and race columns to create independent demographic data
```
smcclure17, line 67: I talked with Nick and he expressed that it would be helpful to have the demographic data in independent entries (instead of combinations of race and age) so this sums over the race and age columns to create individual demographic rows

can_tools/scrapers/official/NH/nh_demographics.py
```diff
@@ -66,7 +66,14 @@ def _get_county(variable: str, county: str):
         )
 
         engine.getWorkbook().setParameter("ShowBy", "Race/Ethnicity")
-        engine.getWorkbook().setParameter("Metric", variable)
+        # By default the parameter is set to "Total Individuals Completed Vaccine"
```
smcclure17, line 69: This scraper is failing on the server but is successful locally. Locally, we were getting some warnings, which I think might be the cause of the failures on the server. 

I don't know why the behavior is different between the two--I ran updates on the Prefect VM to try to fix this, but that didn't change anything

can_tools/utils.py
```diff
@@ -86,3 +86,23 @@ def find_unknown_demographic_id(
 
     df_bad_combinations = df_bad[["age", "race", "ethnicity", "sex"]].drop_duplicates()
     return df_bad_combinations.to_csv(index=False, header=False)
+
+
+def find_duplicated_variable_entries(df: pd.DataFrame, has_location=False):
```
smcclure17, line 91: this detects any duplicated variables. Sometimes scrapers have multiple entries for the same variable (either due to an issue with our scraping or with the dashboard). This helper function highlights those rows (if any exist)

can_tools/bootstrap.py
```diff
@@ -3,9 +3,9 @@
 import sqlalchemy as sa
 from can_tools.models import bootstrap
```
smcclure17, line 5: black reformated this

mikelehen: whoops, thanks.

can_tools/scrapers/official/MI/mi_vaccine_demographics.py
```diff
@@ -63,9 +63,19 @@ def normalize(self, data: pd.DataFrame) -> pd.DataFrame:
...
+        # (e.g 12-15, all, all, all instead of 12-15, ai_an, all, all etc.)
```
mikelehen, line 68: Helpful comment, thanks!

can_tools/scrapers/official/NH/nh_demographics.py
smcclure17: I'm hoping this resolves it, but I'm honestly not sure

mikelehen: Do you know if prefect runs `pip install -e .` or equivalent somehow?  Maybe it is using the old `tableauscraper` dependency?

mikelehen (approved):
Looks good.  I wonder if the prefect server isn't updating it's pip dependencies and so maybe using the old `tableauscraper` dependency or something.

can_tools/scrapers/official/NH/nh_demographics.py
smcclure17: Yeah, that's a good thought. Admittedly, I don't fully know how Prefect pulls in the changes so I'm looking for that workflow/process now. It's just kinda' always worked haha 😭

mikelehen: FWIW- If we can't figure out what Prefect is (or isn't doing) we could log the tableauscraper version at runtime per https://stackoverflow.com/a/4939465/110915 to find out for sure if there's a problem.

smcclure17: Looks like you might be right, if I'm understanding correctly. 

The webhook pulls in the new code and kicks off the Prefect flows here, I think

https://github.com/covid-projections/can-scrapers/blob/ba36f84d9d9806721f9b5f5b2b4ed9e086176f10/services/webhook/update_flows.sh#L3

smcclure17: and it doesn't seem to update the dependencies

smcclure17: ah, didn't refresh to see your comment. Logging seems like a good start

mikelehen: Up to you!  It seems like you're on the right track with that script.  I'm fine with just adding a `pip install -e .` or whatever.

smcclure17: I think I'll add a logger to this to check the version, and if that is the issue then continue looking at that script/the webhooks to try to figure out what's going on, if that sounds good.

mikelehen: 👍  Sounds great.

can_tools/scrapers/official/NH/nh_demographics.py
```diff
@@ -40,6 +43,11 @@ class NHVaccineRace(StateDashboard):
     ]
 
     def fetch(self):
+
+        ts_version = pkg_resources.get_distribution("tableauscraper").version
+        _logger = logging.getLogger(__name__)
+        _logger.warning(f"tableauscraper version {ts_version}")
```
smcclure17, line 49: using a warning so I know it will show up in the cloud logging and since this will be removed shortly

PR #377, opened 2021-11-17
module name TableauScraper --> tableauscraper, aggregate SC data to individual demographics

smcclure17:
Production was using `tableauscraper version 0.1.4` while locally (and in `requirements.txt`) we were using `version 1.29`. 

I upgraded to `version 1.4` locally and attempted to run `SCVaccineDemographics().fetch_normalize()` and got the error that's currently thrown on production. 

I _think_ this might be caused by the case formatting of the tableauscraper module (TableauScraper vs tableauscraper). 
It's a little strange because `pip install tableauscraper` and `pip install TableauScraper` both seem to work.

### DIFF
--- a/can_tools/scrapers/official/SC/sc_demographics.py
+++ b/can_tools/scrapers/official/SC/sc_demographics.py
@@ -38,7 +38,7 @@ class SCVaccineDemographics(StateDashboard):
         ][0]
 
         data = []
-        for county in counties[:2]:
+        for county in counties:
             workbook = engine.setFilter("Recipient County for maps", county)
             for race in self.races:
                 # set the filter functions to select specific race
--- a/can_tools/scrapers/official/SC/sc_demographics.py
+++ b/can_tools/scrapers/official/SC/sc_demographics.py
@@ -38,10 +38,10 @@ class SCVaccineDemographics(StateDashboard):
         ][0]
 
         data = []
-        for county in counties:
+        for county in counties[:2]:
+            workbook = engine.setFilter("Recipient County for maps", county)
             for race in self.races:
-                # set the filter functions to select specific county and race
-                workbook = engine.setFilter("Recipient County for maps", county)
+                # set the filter functions to select specific race
                 workbook = workbook.getWorksheet("Final Age xSex x Race REC")
                 workbook = workbook.setFilter("Assigned Race", race)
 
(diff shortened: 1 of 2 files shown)


can_tools/scrapers/official/SC/sc_demographics.py
```diff
@@ -39,9 +39,9 @@ def fetch(self):
 
         data = []
         for county in counties:
+            workbook = engine.setFilter("Recipient County for maps", county)
```
smcclure17, line 42: don't need to set the county for each race demographic, only when switching counties, so I've moved this outside the inner loop

can_tools/scrapers/official/SC/sc_demographics.py
```diff
@@ -81,3 +81,18 @@ def normalize(self, data: TableauScraper):
             )
             .query("value != -1")
         )
+
+        # sum over age and race columns to create independent demographic data
```
smcclure17, line 85: same deal as with MI, nick preferred to have all the demographics individually

mikelehen (approved):
Looks good, though I didn't quite follow:

> Production was using tableauscraper version 0.1.4 while locally (and in requirements.txt) we were using version 1.29.
> I upgraded to version 1.4 locally ...

It sounds like production was on an old version (0.1.4 vs 0.1.29) so did you _downgrade_ to 0.1.4 locally and that reproduced the error?  Still seems like maybe the fix is to upgrade prod?

smcclure17:
oh, whoops, yeah I downgraded. For some reason, my brain was reading `0.1.29` as `0.1.2.9` so I was thinking 4 > 2.9... 🤦 

You're correct, downgrading to `0.1.4` caused the errors to be thrown, and it works on `0.1.29`

can_tools/scrapers/official/SC/sc_demographics.py
```diff
@@ -81,3 +81,29 @@ def normalize(self, data: TableauScraper):
...
+        # creates dataframes for each demographic that smooths over the other demographics
+        # and concatenate these separate dfs into one.
+        # (e.g create a row like 12-15, all, all, all instead of 12-15, ai_an, all, all etc.)
+        dataframes = []
```
smcclure17, line 89: I realized that this is a bit more complex for this scraper as it gives combinations of age, sex, and race (not just age and race),  so the logic needed to be expanded. 

This feels a little clunky, but I validated the outputs for Aiken County, and the values line up.

PR #378, opened 2021-11-17
Update dependencies on Prefect with webhook 

smcclure17:
#377 did not solve the `tableauscraper` version issue, unfortunately (Production is still using `0.1.4`).

This should hopefully reinstall the dependencies on production whenever code is merged to main. 

I consulted: https://stackoverflow.com/questions/41060382/using-pip-to-install-packages-to-anaconda-environment

### DIFF
--- a/services/webhook/update_flows.sh
+++ b/services/webhook/update_flows.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
 git pull origin main
-sudo apt install python3-pip
-/home/sglyon/miniconda/envs/prefect-can-scrapers/bin/pip3 install -e /home/sglyon/can-scrapers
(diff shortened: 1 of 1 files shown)


services/webhook/update_flows.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
 git pull origin main
+sudo apt install python3-pip
```
smcclure17, line 4: not really sure if we need to install pip here or not

mikelehen (approved):
I think we can probably remove the `apt-install` command, but otherwise looks good!

mikelehen:
From that stackoverflow question, if pip isn't installed in the conda environment then perhaps we'd want to run `conda install pip`...  But I'd just try it as-is to start.

services/webhook/update_flows.sh
smcclure17: good call, I'll remove it and ship it

PR #379, opened 2021-11-17
add requirements to webhook, cleanup NH and SC

smcclure17:
removes the logger from NH, removes an empty location from SC and adds requirement installation to the prefect webhook

### DIFF
--- a/services/webhook/update_flows.sh
+++ b/services/webhook/update_flows.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
 git pull origin main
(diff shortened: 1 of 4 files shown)


services/webhook/update_flows.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
 git pull origin main
+/home/sglyon/miniconda/envs/prefect-can-scrapers/bin/pip install -r /home/sglyon/can-scrapers/requirements.txt
```
smcclure17, line 4: The previous change (to add `pip install -e`) didn't fix the problem, but I was able to resolve the issue by installing the requirements on the VM (via this line of code) so I think this should make it work

mikelehen (approved):

PR #380, opened 2021-11-19
Fix WA demographic scraper dose JSON location keys

smcclure17:
WA scraper data was off by one, causing `unknown` doses to be stored in the `white` bucket, `white` doses in the `ai_an` bucket and so forth: 

context w/ Surgo:

> It looks like our scraper broke over the last couple of days. In short, something changed with the dashboard, and the keys we use to extract each demographic and dose value became off by one causing us to put data in the wrong columns/bins. (So the data in the unknown bucket above is actually the data for the white bucket). The data up until after 11/15 appears to be correct. I’m shipping a quick fix for this right now, and will push the new data through the pipeline. I’ll make changes to the scraper, too, to secure it against this kind of problem in the future.

### DIFF
--- a/can_tools/scrapers/official/WA/wa_vaccine.py
+++ b/can_tools/scrapers/official/WA/wa_vaccine.py
@@ -2,10 +2,7 @@ import asyncio
 
 import pandas as pd
 import us
-import os
 import re
-from pyppeteer import launch
-
 
 from can_tools.scrapers import variables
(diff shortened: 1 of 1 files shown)


mikelehen (commented):
👍  Thanks for taking care of this!

And yeah, I'm curious what kind of sanity checks we could put in place to catch this kind of thing in the future. 🤔

PR #382, opened 2021-12-02
remove NE, fix MN and SC scrapers

smcclure17:


can_tools/scrapers/__init__.py
```diff
@@ -119,14 +119,7 @@
 from can_tools.scrapers.official.NC.nc_vaccine import NCVaccine
 from can_tools.scrapers.official.ND.nd_vaccines import NDVaccineCounty
 
-from can_tools.scrapers.official.NE.ne_vaccines import (
```
smcclure17, line 122: the Nebraska portal now requires a login/authentication :( so our scrapers no longer work. 

The vaccine scrapers are state-level so I've just removed them. I'm leaving the NebraskaCases code here in case they begin to publish data publicly again at some point

can_tools/scrapers/official/SC/sc_demographics.py
```diff
@@ -7,7 +7,7 @@
 from can_tools.scrapers.official.base import StateDashboard
 
 
-class SCVaccineDemographics(StateDashboard):
+class SCVaccineRace(StateDashboard):
```
smcclure17, line 10: SC changed the structure of their dashboard, so this updates the scraper to match (and pulls in ethnicity values)

mikelehen (approved):
Thank you! 🙏

PR #384, opened 2021-12-07
Fix WV, GA, MN and add context to test failures

smcclure17:
Fixes scrapers and pulls in some code from an older branch that gives more context to failing tests. If a test fails due to an unknown demographic, location, or variable value the missing variables/keys are output as seen below. The original error message is still output as well. 

image

### DIFF
--- a/can_tools/utils.py
+++ b/can_tools/utils.py
@@ -2,7 +2,7 @@ from typing import Optional
 import pandas as pd
 import sqlalchemy as sa
 from sqlalchemy.engine.base import Engine
-from can_tools.models import Location, create_dev_engine
+from can_tools.models import create_dev_engine
 
 
 def determine_location_column(df: pd.DataFrame) -> str:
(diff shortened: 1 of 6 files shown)


can_tools/utils.py
```diff
@@ -46,17 +46,25 @@ def load_most_recent_cdc(
     pass
 
 
-def find_unknown_variable_id(df: pd.DataFrame, engine: Engine = None):
```
smcclure17, line 49: update the return types here so these functions can be used in `_create_put_error_msg` below

tests/test_datasets.py
```diff
@@ -95,3 +102,23 @@ def test_covid_dataset_has_source(cls):
...
+    """Find unknown/missing variable entries and format an error message"""
+    unk_demographics = utils.find_unknown_demographic_id(data, engine=engine, csv_rows=True)
+    unk_variables = utils.find_unknown_variable_id(data, engine=engine, csv_rows=True)
+    if cls.location_type in ["county", "state"] or cls.has_location == True:
+        unk_locations = utils.find_unknown_location_id(data, engine=engine, state_fips=cls.state_fips, csv_rows=True)
+    
+    error_msg = (
+        "Exception encountered during scraper put() method. "
+        "This may be due to unknown values in the scraper output. Verify any unknown rows are expected and add them to the corresponding files "
+        "or modify the scraper output to match the expected and existing CSV entries. \n"
```
mikelehen, line 116: Nice! 🎉

mikelehen (approved):

PR #385, opened 2021-12-14
Remove CDCCovidDataTracker, add caching for large scrapers, increase time between MainFlows

smcclure17:
- removes the CDCCovidDataTracker and the Prefect flows related to it
- increases MainFlow schedule from 3 hours to 4 hours
- adds etag checking to large scrapers to only pull in data if it has been updated
- adds a timeout to the `put()` task to fail scrapers if they get stuck


can_tools/scrapers/official/base.py
```diff
@@ -1212,3 +1213,52 @@ def get_dataset(self, body, url) -> str:
...
+
+
+class ETagCacheMixin:
+    """Mixin class to add the ability to check whether a dataset has been updated since last viewed.
+
+    This is used in the `create_cached_flow_for_scraper` flow, to determine whether
+    to ingest the data (execute the fetch, normalize, and put methods).
+    """
+
+    cache_url: str
```
smcclure17, line 1225: We have no standard/universal attribute used to store the URL that the scrapers fetch the data, so I've added one for this class

services/prefect/flows/generated_flows.py
```diff
@@ -73,72 +74,54 @@ def initialize_sentry(sentry_dsn: str):
...
-
-        d.set_upstream(sentry_sdk_task)
-        normalized.set_upstream(fetched)
-        validated.set_upstream(normalized)
-        done.set_upstream(validated)
-
-    return flow
-
-
-def create_cdc_all_states_flow(schedule=True):
```
smcclure17, line 121: Removes the flow that ran the CDCCovidDataTracker flow for all states

services/prefect/flows/generated_flows.py
```diff
@@ -73,72 +74,54 @@ def initialize_sentry(sentry_dsn: str):
...
-        flow.add_task(task)
+        with case(skip_flow, True):
+            skip_cached_flow(cls.__name__)
 
     return flow
 
 
 def create_main_flow(flows: List[Flow], project_name):
-    schedule = CronSchedule("0 */3 * * *")
+    schedule = CronSchedule("0 */4 * * *")
```
smcclure17, line 124: increase the time between MainFlow runs from 3 to 4 hours. The best practice I think would be to have the next MainFlow trigger on the completion of the previous one, but this should be a start

services/prefect/flows/generated_flows.py
```diff
@@ -73,72 +74,54 @@ def initialize_sentry(sentry_dsn: str):
...
+    logger = prefect.context.get("logger")
+    logger.info(f"No new data. Skipping {flow_name}")
+
+
+@task
+def check_if_skip_cached_flow(cls: ETagCacheMixin):
+    return not cls().check_if_new_data()  # if there is new data don't skip this flow.
+
+
 def create_flow_for_scraper(ix: int, cls: Type[DatasetBase], schedule=True):
```
smcclure17, line 88: Modifies `create_flow_for_scraper` to check if the class has etag checking, and if it does and there is no new data we skip the scraper. 

Classes without the `EtagMixinClass` will always run

can_tools/scrapers/cache_scraper_versions.json
```diff
@@ -0,0 +1,18 @@
+{
```
smcclure17, line 1: not sure if this is the best location or format for this file

tests/test_datasets.py
```diff
@@ -1,8 +1,6 @@
 from sqlalchemy.engine.base import Engine
```
smcclure17, line 1: fix a test that included CDCCovidDataTracker

mikelehen (approved):
I think this could use a bit of cleanup if you have time.  Let me know if you want another set of eyes on it after you make changes.

services/prefect/flows/generated_flows.py
```diff
@@ -73,72 +74,54 @@ def initialize_sentry(sentry_dsn: str):
...
 
 
+@task
+def skip_cached_flow(flow_name):
+    logger = prefect.context.get("logger")
+    logger.info(f"No new data. Skipping {flow_name}")
+
+
+@task
+def check_if_skip_cached_flow(cls: ETagCacheMixin):
```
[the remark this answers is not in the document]
smcclure17: good call, I've cleaned this up

can_tools/scrapers/official/base.py
```diff
@@ -1212,3 +1213,52 @@ def get_dataset(self, body, url) -> str:
...
+    cache_name: str
+    cache_file: Path = Path(__file__).parents[1] / "cache_scraper_versions.json"
+
+    @property
+    def etag(self):
+        res = requests.get(self.cache_url)
```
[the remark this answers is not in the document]
smcclure17: Yeah this would've been ideal. Unfortunately, I tried this for the `CDCCountyVaccine2` class and I got the error message:
```
{... 'X-Error-Code': 'invalid_request', 'X-Error-Message': 'HEAD is not supported' ...}
```
using GET is slower, but it should work for all the URLs that we need

can_tools/scrapers/cache_scraper_versions.json
smcclure17: Good points. 

For question 2, I restructured this and now each scraper creates it's own file (and executes the scraper if the file does not exist so that these files don't need to be checked in.)

You make a good point and I agree that it doesn't feel great to be modifying files without any commit history :/

can_tools/scrapers/official/base.py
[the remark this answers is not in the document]
mikelehen: Ah, bummer. Sounds good though.

mikelehen (approved):
This looks great!  Thanks for doing the cleanups.  I like the way this shaped up.

PR #389, opened 2021-12-21
fix VA scraper

smcclure17:
VA scraper data went stale because they restructured their data tables, as noted in https://actnowcoalition.slack.com/archives/C0182MKHWCR/p1640101355000800?thread_ts=1640097831.000600&cid=C0182MKHWCR


mikelehen:
Looks good. Thanks Sean! 👍

PR #390, opened 2021-12-24
replace None with np NaN in CDC testing dataset

smcclure17:
Replacing suppressed values with `None` was not actually replacing the suppressed values (the error `'Unable to parse string "suppressed" at position 0'` was thrown.)

This replaces `None` with `np.NaN` which appears to work.


mikelehen:
👍

PR #391, opened 2022-01-03
Remove duplicates in CDC Testing scraper and fix inheritance in USAFacts scrapers

smcclure17:
The CDC testing dataset now includes a lot of duplicate rows and has multiple values for some (about 65,000) variable entries. Some context is here: https://trello.com/c/9xHrKqo4/137-duplicates-in-cdc-testing-dataset-causing-scraper-to-fail

I've removed the duplicate rows, but I wasn't sure how to best handle the variables with multiple entries. There didn't seem to be a clear indication of which were the "correct" rows, so I've just kept the most recent ones.

Previously `USAFactsDeaths` inherited from `USAFactsCases` so when `super().__init__()` was called, it would call the init of `USAFactsCases`, which would call `EtagCacheMixin.initialize_cache()` inside the cases scraper object. This caused the deaths scraper to have the same `cache_file` as the cases scraper. This restructures things so that the cases and deaths scraper inherit from an abstract base class, removing this issue.


can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
```diff
@@ -74,3 +74,11 @@ def normalize(self, data: pd.DataFrame) -> pd.DataFrame:
                 locations_to_drop=[2066, 2063, 72888, 72999],
             ).pipe(self._reshape_variables, variable_map=self.variables)
         )
+
+        # There are duplicated rows and multiple rows for a single variable in the dataset.
+        data = data.drop_duplicates()
+        # If a variable (e.g. a location, date, and variable type) has multiple entries just select the most
+        # recent occurrence. There does not seem to be a pattern or a discernably "correct" value.
```
mikelehen, line 81: How confident are we that this will work out to give us valid data?  Should we proactively build a snapshot with this and validate it today before tomorrow's deploy?

mikelehen (approved):
Looks good!

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
smcclure17: I'm fairly confident it will be valid--the values I saw seemed to be pretty close--but good idea, definitely worth building a snapshot before tomorrow

PR #394, opened 2022-01-10
add state HHS testing and NYT scrapers

smcclure17:


can_tools/scrapers/nytimes/nyt_cases_deaths.py
```diff
@@ -0,0 +1,274 @@
...
+            .pipe(
+                self._reshape_variables,
+                variable_map=self.variables,
+                id_vars=["location_type"],
+            )
+            # two Alaska FIPS not in our location csv
+            .query("location not in [2997, 2998]")
+        )
+
+
```
smcclure17, line 169: Everything from here and below is effectively copied from [covid-data-public](https://github.com/covid-projections/covid-data-public/blob/main/scripts/update_nytimes_data.py)

can_tools/scrapers/nytimes/nyt_cases_deaths.py
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    variables = {
+        "cases": CUMULATIVE_CASES_PEOPLE,
+        "deaths": CUMULATIVE_DEATHS_PEOPLE,
+    }
+
+    def __init__(self, execution_dt: pd.Timestamp = pd.Timestamp.utcnow()):
+        # All the files appear to be updated at the same time.
+        # Use the national file for cache checking since it's the smallest
```
mikelehen, line 126: Hrm. I'm not sure if we should count on this.  It's probably true for the daily deploy but if they post mid-day updates I'm not sure it'll be true, and I could see this causing difficult-to-debug confusion down the road.

[Here](https://github.com/nytimes/covid-19-data/commit/d50443c374d8b614cc386667a0a63d11fe77740f)'s an example where /us-counties.csv was updated but /us.csv was not.

Maybe we should just not bother doing caching for NYT?  Or will that cause meaningful problems you think?  Otherwise I guess we have to turn it into 3 different scrapers or come up with a different way to do caching?

mikelehen (approved):
Looks great! Thank you!

can_tools/scrapers/nytimes/nyt_cases_deaths.py
smcclure17: ah that's true, good catch. 

Initially (in a9070ecb0828968dae8f7a64ef7f3de0a7176f30), I had these split up into separate scrapers, but I merged them together because the backfill code modifies both the state and county data.

Let's try this without caching for the time being, and if that does cause problems we can revisit.

PR #395, opened 2022-01-11
Add CDC booster data and update vaccine schema (and rename CDC scraper)

smcclure17:
The [CDC booster column](https://data.cdc.gov/Vaccinations/COVID-19-Vaccinations-in-the-United-States-County/8xkx-amqh) is documented as: 
```Total number of people who are fully vaccinated and have received a booster (or additional) dose.```
which suggests that if someone were to receive a second booster it would not be counted in this column. I'm not sure how the CDC plans to track 4th-shot-boosters when they are authorized (if they will update this column to include those doses or create a new column etc.)

At the moment I've created the CMU variable to mirror the CDC description, but I wonder if we should track `doses` instead of `people` so that we can continue to use this variable when future doses are authorized (e.g. by combining future 3rd-, 4th- and subsequent-shot columns into one).

### DIFF
--- a/can_tools/bootstrap_data/covid_categories.csv
+++ b/can_tools/bootstrap_data/covid_categories.csv
@@ -67,6 +67,7 @@ vaccine,total_vaccine_allocated
 vaccine,total_vaccine_distributed
 vaccine,total_vaccine_initiated
 vaccine,total_vaccine_completed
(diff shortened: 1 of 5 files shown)


mikelehen (approved):
Nice!  That was pretty easy I guess. Excellent!

PR #401, opened 2022-01-18
add CDC "As originally posted" Testing data

smcclure17:
This also makes some tweaks to use the `autodag` class variable to ignore certain classes (in this case the `CDCTestingBase` class)

### DIFF
--- a/can_tools/__init__.py
+++ b/can_tools/__init__.py
@@ -35,20 +35,10 @@ def all_subclasses(cls):
     )
 
 
-def non_abc_subclasses(cls):
-    out = list(x for x in all_subclasses(cls) if not inspect.isabstract(x))
-
-    # need to specifically remove some classes :(
-    from can_tools.scrapers.official.base import TableauDashboard
-    from can_tools.scrapers.official.PA.pa_vaccines import (
-        PennsylvaniaVaccineDemographics,
-    )
-
-    bad = [TableauDashboard, PennsylvaniaVaccineDemographics]
-
-    return [x for x in out if x not in bad]
+def scrapers_for_flow(cls):
+    return list(x for x in all_subclasses(cls) if not inspect.isabstract(x) and cls.autodag)
 
 
-ALL_SCRAPERS: List[Type[scrapers.DatasetBase]] = non_abc_subclasses(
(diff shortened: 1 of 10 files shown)


can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
```diff
@@ -39,15 +39,15 @@ class CDCHistoricalTestingDataset(FederalDashboard, ETagCacheMixin):
...
+            self, cache_url=self.fetch_url, cache_file=self.cache_file
         )
         super().__init__(execution_dt=execution_dt)
 
     def fetch(self) -> pd.DataFrame:
         # select only the columns we care about in order to speed up query
         data = pd.read_json(
             f"{self.fetch_url}?$limit={SODA_API_RESPONSE_LIMIT}"
-            "&$select=fips_code,date,cases_per_100k_7_day_count,percent_test_results_reported"
+            f"&$select=fips_code,{self.date_column},cases_per_100k_7_day_count,percent_test_results_reported"
```
smcclure17, line 50: The two datasets have different names for the data column, so we have to specify

can_tools/__init__.py
```diff
@@ -35,20 +35,10 @@ def all_subclasses(cls):
     )
 
 
-def non_abc_subclasses(cls):
-    out = list(x for x in all_subclasses(cls) if not inspect.isabstract(x))
-
-    # need to specifically remove some classes :(
```
smcclure17, line 41: Instead of excluding these scrapers here, I've disabled them using `autodag = False` in the scrapers themselves. 

The Tableau class is a parent class so it shouldn't be included, but I don't have context as to why the PA scraper is disabled

mikelehen: Hrm, yeah.  Not 100% sure. I remember there was a discrepancy with Philadelphia in the PA state dashboard, so maybe we just disabled the whole scraper?  Not sure.

can_tools/scrapers/official/federal/CDC/cdc_testing_cases.py
mikelehen: LOL, of course they do. 🙄

mikelehen (approved):
LGTM.

PR #402, opened 2022-01-20
fix Philadelphia scraper and switch CDC state scraper to timeseries data

smcclure17:
Fixes the PA scraper and switches the state CDC scraper from their API to their official dataset, which gives us timeseries data.

### DIFF
--- a/can_tools/scrapers/official/PA/philadelhpia_vaccine.py
+++ b/can_tools/scrapers/official/PA/philadelhpia_vaccine.py
@@ -29,17 +29,17 @@ class PhiladelphiaVaccine(TableauDashboard):
 
     def fetch(self) -> pd.DataFrame:
         # create a dict of the 2 dose type tables
-        # which are titled "Residents Percentage New" and "... Full"
+        # which are titled "Residents Percentage First" and "... Full"
         return {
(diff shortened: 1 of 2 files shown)


can_tools/scrapers/official/PA/philadelhpia_vaccine.py
```diff
@@ -29,17 +29,17 @@ class PhiladelphiaVaccine(TableauDashboard):
...
         # create a dict of the 2 dose type tables
-        # which are titled "Residents Percentage New" and "... Full"
+        # which are titled "Residents Percentage First" and "... Full"
         return {
             dose_type: self.get_tableau_view(dose_type=dose_type)[
-                self.data_tableau_table.format(dose_type=dose_type)
+                self.data_tableau_table.format(dose_type=table_name)
             ]
-            for dose_type in ["New", "Full"]
+            for dose_type, table_name in {"First": "New", "Full": "Full"}.items()
```
smcclure17, line 37: This is pretty clunky but, previously the names of the dashboards matched the names of the tables inside them that we wanted to pull from. The names of the tables changed and don't quite match anymore, so this modifies the code to work while keeping the same type of structure

mikelehen (approved):

PR #405, opened 2022-01-28
Revert autodag scraper filtering

smcclure17:
My change in #401 to filter out scrapers with `autodag` set to false caused issues. 

Setting `autodag = False` for `TableauDashboard` also set it to false for all of its subclasses, meaning none of them (about 12 in total, including the North Carolina scraper) were being included in the `MainFlow`s 😬. 

I'm sorry about this, I'm not sure how I didn't realize this at the time, but I'll do more testing before making similar changes in the future. Fortunately, this didn't affect any scrapers that we are currently using, but it does mean that we lost ~10 days of history for the scrapers that were filtered out. This reverts the filtering to the old method (explicitly removing the scrapers during the collection).

### DIFF
--- a/can_tools/__init__.py
+++ b/can_tools/__init__.py
@@ -27,6 +27,9 @@ import inspect
 from typing import List, Type
 
 from can_tools import scrapers
+from can_tools.scrapers.official.base import TableauDashboard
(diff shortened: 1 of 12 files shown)


mikelehen (approved):
Whoops!  No worries. I could have caught that in code review too if I'd been a bit more diligent. Thanks for tracking down and fixing.

PR #407, opened 2022-02-01
Add NC state-level vaccine scraper

smcclure17:
This also adds booster data to the county-level scraper. We should be able to get historical booster data for the counties by using the [rerun_scraper](https://github.com/covid-projections/can-scrapers/blob/ba36f84d9d9806721f9b5f5b2b4ed9e086176f10/can_tools/scraper_maintenance.py#L42) functionality (running the `normalize()` method on the saved outputs of the previous `fetch()` runs from GCS), though, I've never actually tried it before. 

This, unfortunately, won't be possible for the state data as it comes from different tables, so we'll have no historical data.


mikelehen (approved):
Looks great!

PR #408, opened 2022-02-02
Add empty insert exception and timout for Update Parquet Files task

smcclure17:


can_tools/scrapers/official/base.py
```diff
@@ -120,6 +120,12 @@ def _put_exec(self, engine: Engine, df: pd.DataFrame) -> None:
...
                 rows_deleted = res_delete.rowcount
                 print("Removed the {} rows from temp table".format(rows_deleted))
+            
+        if rows_inserted == 0:
+            raise InsertedZeroRowsError(
+                "_put_exec() method completed successfully "
+                "but inserted zero rows into database."
+            )
 
         return worked, rows_inserted, rows_deleted
```
smcclure17, line 130: We have this `worked` variable that is supposed to return `True` or `False` depending on the outcome of this method. We [don't ever do anything with it](https://github.com/covid-projections/can-scrapers/blob/main/services/prefect/flows/generated_flows.py#L66-L67), and I think it will only return ever return `True` because (AFAICT) if an exception is thrown the method will terminate. 

So, I found it easiest just to raise an exception in this method directly

mikelehen: Looks like the code above uses `fast_append_to_sql()` to write the data from the dataframe into a temporary table and then the actual insert is copying the data over from there.  So I wonder if there's an (easy) way to make `fast_append_to_sql()` log the number of rows it inserted into the temporary table (or anything else that'd be useful).  That way if this happens again we'll which part failed. 🤷 

Food for thought. Doesn't need to be in this PR.

can_tools/scrapers/official/base.py
```diff
@@ -121,6 +121,12 @@ def _put_exec(self, engine: Engine, df: pd.DataFrame) -> None:
                 rows_deleted = res_delete.rowcount
                 print("Removed the {} rows from temp table".format(rows_deleted))
 
+        if rows_inserted == 0:
```
mikelehen, line 124: Are we confident that we never intentionally write 0 rows?  If we were worried about that we could add a check for `df.shape[0] == 0` here or at the top of the method or something.

mikelehen (approved):
Looks good!

can_tools/scrapers/official/base.py
smcclure17: agreed, that would give us some more insight. I think I'm going to ship this as is, but can add that to the backlog

can_tools/scrapers/official/base.py
smcclure17: I'm not 100% confident, but it seems that we should never need to. I've extended this `if` in f66c34f53240642436abab23cb2542357b461f0c to not throw the error if `df` is actually empty

PR #410, opened 2022-02-16
Create dedicated flow for updating NYT data and triggering parquet update

smcclure17:
Creates a flow to check the NYT data source every 30 minutes, and on new data trigger the update parquet file flow. This follows the workflow outlined in the docs: https://docs.prefect.io/core/idioms/flow-to-flow.html


services/prefect/flows/nyt_updater.py
```diff
@@ -0,0 +1,34 @@
...
+
+from prefect.tasks.prefect import create_flow_run, wait_for_flow_run
+from prefect.utilities.notifications import slack_notifier
+from prefect.engine.state import Failed
+
+
+def init_updater_flow():
+    with Flow(
+        "NYTParquetUpdater",
+        schedule=CronSchedule("30 6 * * *"),
```
mikelehen, line 12: Can you add a comment something like

> NYT typically publishes their daily data update at 1:10AM EST and so we kick off our scraper at 1:30AM EST.

As an aside, I wonder if we're going to have any issues with getting "desynched" from NYT when daylight savings time ends...

I wonder if we should be thinking about trying to implement something like we had in covid-data-public where we just query their repo every 30min to see if they've pushed data.  That way we aren't reliant on this timing working out.

services/prefect/flows/nyt_updater.py
```diff
@@ -0,0 +1,34 @@
+from prefect import Flow
```
mikelehen, line 1: `nyt_updater.py` feels too generic.  I wonder if it should be scheduled_nyt_and_parquet_updater.py or something? 

Maybe also add a comment explaining that this is the main updater that is responsible for getting NYT data into a parquet file in time for the snapshot build to pick it up.

services/prefect/flows/nyt_updater.py
```diff
@@ -0,0 +1,34 @@
...
+
+def init_updater_flow():
+    with Flow(
+        "NYTParquetUpdater",
+        schedule=CronSchedule("30 6 * * *"),
+        state_handlers=[slack_notifier(only_states=[Failed])]  # notify on failure only,
+    ) as flow:
+
+        # NYTimesCasesDeaths and UpdateParquetFiles flows must already be registered via
+        # generated_flows.py and update_api_view.py
```
mikelehen, line 17: Not 100% sure what to make of this comment.  I assume if they're not, the code below will just fail?  I could imagine a more useful comment being something like:

> Note that the below code relies on NYTimesCasesDeaths and UpdateParquetFiles flows already being registered via generated_flows.py and update_api_view.py.  This is guaranteed by ...?

But I'm not actually sure what's guaranteeing it!  What controls the order of registration?  What are we doing to ensure they're registered in the right order?

services/prefect/flows/nyt_updater.py
```diff
@@ -0,0 +1,34 @@
...
+        nyt_flow = create_flow_run(
+            flow_name="NYTimesCasesDeaths", project_name="can-scrape"
+        )
+        wait_for_nyt = wait_for_flow_run(nyt_flow, raise_final_state=True)
```
mikelehen, line 21: Out of curiosity, what would happen if NYT was already running and so when we try to create the flow here it ends up skipping it?

Would we immediately go ahead and create the parquet file (without up-to-date NYT data)?  Or would this flow fail?  Or... what?

services/prefect/flows/nyt_updater.py
```diff
@@ -0,0 +1,34 @@
...
+
+        parquet_flow = create_flow_run(
+            flow_name="UpdateParquetFiles", project_name="can-scrape"
+        )
+        wait_for_parquet = wait_for_flow_run(parquet_flow, raise_final_state=True)
+
+        parquet_flow.set_upstream(wait_for_nyt)
```
mikelehen, line 28: Not 100% sure what this `set_upstreame()` is for, but wondering if it should be `wait_for_parquet` instead of `wait_for_nyt`?

mikelehen (approved):
This looks good but left a bunch of minor comments.

I'm also a bit unsure of how this fits into our existing flows?  Don't we already have a NYT scraper run scheduled for this time?  I would kind of expect to see a diff in here that disables that one in favor of this one or something.  I'm probably just missing some context though.

services/prefect/flows/nyt_updater.py
smcclure17: Good question--I'm not sure... I'll look in the docs/do some testing to find out. 

Ultimately, we would probably want to cancel the existing instance and have this one run in case the previous run had started before the data was updated (but I'm not immediately sure how that might be done).

services/prefect/flows/nyt_updater.py
smcclure17: I believe what `set_upstream()` does here is make the `parquet_flow` flow wait until `wait_for_nyt` has been completed before executing. `wait_for_nyt` is a task that waits to complete until the `nyt_flow` has finished. I think basically what's it's saying is "wait for the NYT flow to finish before running the Parquet flow".

So, I think that here the `wait_for_parquet` variable actually isn't needed, but it should still be `parquet_flow.set_upstream(wait_for_nyt)`. 

The docs ([here](https://docs.prefect.io/api/latest/core/task.html#task-2) and [here](https://docs.prefect.io/api/latest/tasks/prefect.html#wait-for-flow-run)) give a little bit more context, but they don't give a huge amount of explanation

mikelehen: Oh!  I see, yeah, that makes sense.  I got confused because of the ordering (calling `wait_for_flow_run(parquet_flow, ...)` before specifying that the parquet_flow depends on the nyt_flow)...

services/prefect/flows/nyt_updater.py
smcclure17: Yeah, I should make this more clear. 

Flows only need to be registered once per project, and once they've been registered they can be called/run whenever.  For most of the scraper flows registration happens in [`generated_flows.py`](https://github.com/covid-projections/can-scrapers/blob/main/services/prefect/flows/generated_flows.py#L151-L153), and for the update parquet flow it happens in [`update_api_view.py`](https://github.com/covid-projections/can-scrapers/blob/main/services/prefect/flows/update_api_view.py#L59-L66). So while it's not guaranteed, as long as those files have been run in the past (via `python update_api_view.py` or something) these flows should exist.

services/prefect/flows/nyt_updater.py
smcclure17: Yeah, I agree that it would be ideal to constantly be checking the NYT data

smcclure17:
Thanks for this feedback 🙏, I'll work to revise this into something a bit more clean/clear. 

Regarding, 
> Don't we already have a NYT scraper run scheduled for this time? 

Good catch. You're able to add schedules to existing flows in the Prefect dashboard/UI without making any code changes (by going to a specific flow > settings > new schedule), so this is how I added that schedule for the NYT scraper. 

We'll want to remove that schedule whenever we merge this PR.

services/prefect/flows/nyt_updater.py
smcclure17: (I'll look to add this during my revisions)

mikelehen:
Ah!  That makes sense.  Out of curiosity, do you know if there's any way to see what flows are set up to run on a schedule?  Like assuming I came to our dashboard not knowing anything, is there any way I could have found out that the NYT scraper runs on a schedule other than by clicking every single flow and going to its settings?  Guessing not (couldn't immediately find something in Google), but thought I'd ask.

smcclure17:
I don't think there's anywhere that explicitly lists out the flow schedules, no :(

I think the best way to check is to look in the "Upcoming Runs" section, which lists out the runs that are currently scheduled. Scrolling through here you would be able to see that the NYT scraper is scheduled every day at 1:30 ET, the MainFlows run every 4 hours, etc., but it's not perfect
 
image

services/prefect/flows/scheduled_nyt_and_parquet_updater.py
```diff
@@ -0,0 +1,53 @@
...
+            task_args=dict(name="run NYTimesCasesDeaths flow"),
+        )
+
+        # Execute on success, skip on skipped result of wait_for_nyt
+        parquet_flow = create_flow_run(
+            flow_name="UpdateParquetFiles",
+            project_name="can-scrape",
+            upstream_tasks=[wait_for_nyt],
+            task_args=dict(
+                name="Create UpdateParquetFiles flow", skip_on_upstream_skip=True
```
smcclure17, line 38: With `skip_on_upstream_skip=True` if `wait_for_nyt` finishes in a skipped state, this flow (the parquet flow) will also be skipped.

can_tools/scrapers/nytimes/nyt_cases_deaths.py
```diff
@@ -120,6 +120,15 @@ class NYTimesCasesDeaths(FederalDashboard):
         "deaths": CUMULATIVE_DEATHS_PEOPLE,
     }
 
+    def __init__(self, execution_dt: pd.Timestamp = pd.Timestamp.utcnow()):
```
smcclure17, line 123: add caching for the NYT scraper so that we can check the source every 30 minutes. This uses https://api.github.com/repos/nytimes/covid-19-data/commits as the URL due to [previous concerns](https://github.com/covid-projections/can-scrapers/pull/394#discussion_r782535124), which I think should get a new ETag whenever a new commit is made

services/prefect/flows/generated_flows.py
```diff
@@ -75,9 +77,13 @@ def initialize_sentry(sentry_dsn: str):
 
 
 @task
-def skip_cached_flow(flow_name):
-    logger = prefect.context.get("logger")
-    logger.info(f"No new data. Skipping {flow_name}")
+def skip_cached_flow():
```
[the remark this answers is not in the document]
mikelehen: Thanks!  As an aside, whenever you write a long comment on a github code review it's probably worth asking if part(s) of it should be in a comment directly in the code instead.

In this case, the code comment is pretty good, but explicitly referencing `etag_caching_terminal_state_handler` instead of just "The flow's terminal state handler" might be helpful to somebody coming along later and trying to understand this code.

services/prefect/flows/utils.py
```diff
@@ -39,3 +41,35 @@ def skip_if_running_handler(obj, old_state, new_state):
...
+    """Update flow's final state to Skipped if a reference task is skipped with etag_skip_flag = True.
+
+    This is used in conjunction with the skip_cached_flow() task to manually skip a
+    flow if the Etag caching does not detect new data. This allows us to conditionally
+    run flows downstream of a scraper flow, depending on whether or not there is new data.
+    This conditional logic is used in the NYTParquetUpdater flow in
+    scheduled_nyt_and_parquet_updater.py.
+
+    For more info on terminal state handlers:
+        https://docs.prefect.io/core/concepts/flows.html#terminal-state-handlers
```
mikelehen, line 60: Super helpful comment, thank you! 🙏

services/prefect/flows/utils.py
```diff
@@ -39,3 +41,35 @@ def skip_if_running_handler(obj, old_state, new_state):
...
+    """
+
+    # look through all the flow's tasks to see if any we're skipped.
```
mikelehen, line 63: nit: we're => were

services/prefect/flows/utils.py
```diff
@@ -39,3 +41,35 @@ def skip_if_running_handler(obj, old_state, new_state):
...
+    # If so, and the etag_skip_flag attribute is True, set the final state of the flow to Skipped.
+    # etag_skip_flag is a custom flag passed to a task's skip signal during the etag cache checking process.
+    # Any tasks that normally finish with the Skipped state will not trigger this state handler.
+    for task_state in reference_task_states:
+        if task_state.is_skipped():
+            if getattr(task_state, "etag_skip_flag", False) is True:
+                return Skipped(
+                    "Setting final state to skipped due to reference task"
+                    "skipped with attribute etag_skip_flag = True"
```
mikelehen, line 72: nit: Do we need a space to avoid this becoming `... taskskipped ...` ?

mikelehen (approved):
Left a couple nits, but I think this looks great!  Thanks for figuring out the skipped task / flow plumbing stuff and the NYT caching.

services/prefect/flows/generated_flows.py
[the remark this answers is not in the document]
smcclure17: Good point, and I've added a reference to etag_caching_terminal_state_handler in 629ede449fdf370211f6e549ce4c1915c8c378e1

services/prefect/flows/utils.py
smcclure17: we do yeah 🙏

PR #412, opened 2022-02-16
Only allow one instance of each flow to run at a time

smcclure17:
Adds a state handler that, whenever a flow run is set to execute, checks to see if the current flow already has another instance running and if so, skips the instance that's about to start. 

Last night, it looks like two NYT scrapers kicked off around the same time (see screenshots below), and I think this might be what caused them to take longer to complete. #410 eliminates the need to rely on scheduling/timing, so if the scrapers take longer in the future it won't be as big of a deal, but regardless, there's no need to have two instances of the same scraper flow running at the same time. 

image
image

### DIFF
--- a/services/prefect/flows/utils.py
+++ b/services/prefect/flows/utils.py
@@ -2,13 +2,14 @@ import prefect
 from prefect.client import Client
 from prefect.engine.state import Skipped
 
+
 def skip_if_running_handler(obj, old_state, new_state):
     """State handler to skip flow if another instance is already in progress.
     
     see: https://github.com/PrefectHQ/prefect/discussions/5373
     """
     logger = prefect.context.get("logger")
-    
+
     # queries the graphql server to see if any instances of the same flow
     # is already running, and if so, sets the flow state to Skipped
     if new_state.is_running():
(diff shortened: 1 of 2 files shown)


services/prefect/flows/utils.py
```diff
@@ -0,0 +1,38 @@
...
+                start_time
+              }
+            }
+        """
+        response = client.graphql(
+            query=query, variables=dict(flow_id=prefect.context.flow_id)
+        )
+        active_flow_runs = response["data"]["flow_run"]
+        if active_flow_runs:
+            message = "Flow already has a run in progress, skipping..."
```
mikelehen, line 35: I wonder if it'd be worth including `response["data"]["start_time"]` in this log message.  That way if we end up with issues of flows being skipped because a previous instance is running too long or something it'll be easier to diagnose.

mikelehen (approved):
Looks reasonable.

Any idea if we should be worried about issues with flows getting hung, leading to subsequent runs being skipped, and us not noticing for a while?  Do we have timeouts on most flows?  I'm not too worried about it.  Just flagging it as a potential issue to watch out for.

smcclure17:
Yeah, that's a good flag, I think we maybe should be. We get the start time of the currently running flow from the request, so we could potentially just say "if the running flow has been running for 2 hours, start anyway" but that's hacky. 

The best route is to make sure we have proper timeouts for all the flows, which I'm not sure we do at the moment.

PR #413, opened 2022-02-16
Trigger covid-data-model Update Combined Datasets action from Prefect flow

smcclure17:
Add a flow to run after UpdateParquetFiles to trigger the covid-data-model Update Combined Datasets action via an HTML request. In the case that there is no new data, this flow will not execute. 

This also adds a slack notifier to ping #dev-alerts whenever NYTParquetScheduledUpdater fails.  

This [update datasets run](https://github.com/covid-projections/covid-data-model/runs/5239855891?check_suite_focus=true) was triggered by the contents of `make_request()`

Post-merge TODOs:

- [ ] Remove the Cron schedule from Update Combined Datasets in covid-data-model


services/prefect/flows/scheduled_nyt_and_parquet_updater.py
```diff
@@ -1,17 +1,52 @@
...
+        headers={
+            "Authorization": f"token {github_token}",
+            "Accept": "application/vnd.github.v3+json",
+        },
+        data='{"ref": "main"}',
+    )
+    response.raise_for_status()  # raise status in case of failure
+
+
+def trigger_data_model_flow():
```
mikelehen, line 29: Nit: Do we have a naming convention for naming flows?  Seems a bit confusing that `init_scheduled_nyt_updater_flow()` is prefixed with `init_` but this one isn't... But maybe there's a difference I'm not understanding.

mikelehen (approved):
Looks good with tiny nit.

services/prefect/flows/scheduled_nyt_and_parquet_updater.py
smcclure17: We don't, yeah, we should make this more uniform. They both do the same thing--I'll update this one to match `init_scheduled_nyt_updater_flow` when I get back to this

PR #414, opened 2022-02-18
Move etag_skip_flag to signal's context to fix terminal state handling. 

smcclure17:
Originally I thought we could just add arbitrary tags/arguments to a state signal. The code I had in place before just created a signal message that looked like: `"{'message': 'No new source data. Skipping...', 'etag_skip_flag': True}"` (note that it's just one big string 🤦). So, there never was an `etag_skip_flag` attribute to look for, meaning the final flow state would never be set to `Skipped`. 

To fix this, this now modifies the [signal's context](https://github.com/PrefectHQ/prefect/blob/4e9cc138e98535e45937e53c972f718af5144f60/src/prefect/engine/state.py#L41) and adds the `etag_skip_flag` to this dictionary. 

My local testing must have been slightly different to not catch this :/

### DIFF
--- a/services/prefect/flows/generated_flows.py
+++ b/services/prefect/flows/generated_flows.py
@@ -78,11 +78,11 @@ def initialize_sentry(sentry_dsn: str):
 
 @task
 def skip_cached_flow():
-    # Set the task state to skipped.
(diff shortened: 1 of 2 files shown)


mikelehen (commented):
👍 Thanks for the follow-up and explanation.

PR #415, opened 2022-02-22
Use PrefectSecret instead of EnvVarSecret for MonitorFailingScrapers flow

smcclure17:
This should revive the `MonitorFailingScrapers`. 

In order to add slack notifying in #413, we had to set the slack webhook URL as a [Prefect Secret](https://docs.prefect.io/orchestration/concepts/secrets.html#setting-local-secrets) (I used the `~/.prefect/config.toml` method). This is the same URL used for the  `MonitorFailingScrapers` flow, and accessing this secret instead of using an environment variable seems to work (although I'm still not sure why `EnvVarSecret` isn't working).

### DIFF
--- a/services/prefect/flows/monitor_failing_scrapers.py
+++ b/services/prefect/flows/monitor_failing_scrapers.py
@@ -7,7 +7,7 @@ import requests
 import prefect
 from prefect import Flow, task
 from prefect.schedules import CronSchedule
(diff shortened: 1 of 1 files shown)


mikelehen (approved):
Weird, but looks good! 🤷  Thanks for getting this figured out!

PR #416, opened 2022-02-23
Fix WA, VT vaccine scrapers, remove WI case scrapers

smcclure17:
Removing WI state and county case scrapers since we no longer scrape case data from state dashboards.


mikelehen (approved):
👍

PR #419, opened 2022-03-14
Temporarily add USAFacts to NYT scheduled runs

smcclure17:
The goal here is to ping the USA Facts endpoint every 30 minutes to understand when in the day the data is typically updated. I have just added this into the existing NYT scheduled updater. 

If the USA Facts data is updated, it will not trigger the parquet updater or the Github action like the NYT flow does. 

If we're ok with merging this, I'll add a reminder to my calendar to remove this on Thursday.


mikelehen (approved):
Ah, nice. Thanks for doing this!

PR #420, opened 2022-03-14
Add hospital admissions to existing HHS scrapers

smcclure17:
Add total new 7 day county-level covid hospital admissions and new daily state-level admissions to HHS scrapers

State data source: https://healthdata.gov/Hospital/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/g62h-syeh
County data source: https://healthdata.gov/Hospital/COVID-19-Reported-Patient-Impact-and-Hospital-Capa/anag-cw7u/


can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -105,6 +105,21 @@ def normalize(self, data: str) -> pd.DataFrame:
         temp = temp + df.eval(f"{pps} / {ppc}").fillna(0.0)
         df.loc[:, "inpatient_beds_used_covid_7_day_canavg"] = temp.values
 
+        # Calculate adult and pediatric hospital admissions (same process as above).
+        adult_admissions = "previous_day_admission_adult_covid_confirmed_7_day_sum"
```
smcclure17, line 109: I've copied the averaging done above here, but is it possible that we might want to instead collect the raw `sum` of admissions, to try to match the CDC's indicator of `New COVID-19 admissions per 100,000 population (7-day total)`?

mikelehen: Hrm, interesting...  I *think* this code makes sense as-is, but we'll definitely want to at minimum sanity-check it against the CDC data.

My understanding is that hospitals have a "survey" to fill out, ideally daily, and one of the questions is basically "How many confirmed covid admissions did you have in the previous day?" and the _sum column adds those answers up over 7 days. BUT they might not fill it out every day, so the _coverage number could be 0-7 indicating how many times they actually filled it out.  To get a valid estimate for the week, I think we *have* to take sum/coverage, else we would end up with lower sums for hospitals that report less frequently.  

I don't 100% trust myself though (notably the metrics we did this for before were "current" metrics, not "new" metrics so having <7 samples is less of a problem).  So please do try to sanity-check this in any way that makes sense to you.

can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -105,6 +105,21 @@ def normalize(self, data: str) -> pd.DataFrame:
...
+        adult_coverage = "previous_day_admission_adult_covid_confirmed_7_day_coverage"
```
mikelehen, line 110: Out of curiosity, do we know for sure if CDC is using "confirmed" or "suspected" data?

mikelehen (approved):
Looks good!

can_tools/scrapers/official/federal/HHS/facility.py
smcclure17: It looks like they use confirmed data, as that's what's specified in the footnotes here: https://covid.cdc.gov/covid-data-tracker/#county-view

can_tools/scrapers/official/federal/HHS/facility.py
smcclure17: I see, thanks for this context 🙏 🙏 

It seems that the CDC must be using the `_sum` data, or some variation of it based on their values. I'm not sure if they're concerned about hospitals not reporting every day (and if they aren't, why not) or if they are accounting for this somehow.

Our values do not quite line up 1:1 even with this disaggregation (though they are usually close), so I'm in the process of hashing out why that might be.

can_tools/scrapers/official/federal/HHS/facility.py
```diff
@@ -105,6 +103,15 @@ def normalize(self, data: str) -> pd.DataFrame:
         temp = temp + df.eval(f"{pps} / {ppc}").fillna(0.0)
         df.loc[:, "inpatient_beds_used_covid_7_day_canavg"] = temp.values
 
+        # Combine adult and pediatric covid admissions to find total.
+        # NOTE: Theoretically, this could be missing data from the weekly total
+        # if hospitals do not report everyday, but this seems to be the methodology that
+        # the CDC is following for weekly hospital admissions
```
mikelehen, line 109: For my own understanding, how can you tell what methodology CDC is using?

I think in practice it doesn't matter because looking at the data, 95% of hospitals have _coverage of 7 and only 0.5% of hospitals have _coverage<7 and _admissions>0 (in the past week anyway, might be different during a surge).  But based on my understanding of these columns, the most correct data would be `(_sum/_coverage)*7` to get an approximated 7-day total.  But we're literally talking about 26 facilities and most of them have coverage of 6, so it wouldn't change the numbers that much anyway.

Can you adjust the NOTE to be clearer about our rationale here?  And if the rationale is "We think the CDC is using the raw _sum and only 0.5% of hospitals have _coverage<7 and _sum>0 so it doesn't matter much." that would be fine.

mikelehen (approved):

PR #423, opened 2022-04-04
Fix deploy airflow workflow.

mikelehen:
The last two deploy airflow jobs failed with:

> Error: On 2022-04-05, the default branch will be renamed from "master" to "main". Your action is currently pinned to "@master". Even though GitHub creates redirects for renamed branches, testing found that this rename breaks existing GitHub Actions workflows that are pinned to the old branch name.
>
> We strongly advise updating your GitHub Action YAML from:
>
>     uses: 'google-github-actions/setup-gcloud@master'
>
> to:
>
>     uses: 'google-github-actions/setup-gcloud@v0'

So this updates it with the fix.  Successful run [here](https://github.com/covid-projections/can-scrapers/actions/runs/2087279165)

### DIFF
--- a/.github/workflows/deploy-airflow.yml
+++ b/.github/workflows/deploy-airflow.yml
@@ -19,7 +19,7 @@ jobs:
       FORCE_COLOR: 1
     steps:
     - uses: actions/checkout@v2
-    - uses: google-github-actions/setup-gcloud@master
+    - uses: google-github-actions/setup-gcloud@v0
       with:
         version: '318.0.0'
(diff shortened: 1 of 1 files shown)


smcclure17 (approved):
Sweet, thanks for this fix,

FWIW, I don't think we actually use Airflow anywhere anymore, so we could probably get rid of this action if we feel like it. In general, it might be nice to rework all the tests in this repo as they take about 40 minutes to run and never pass, (maybe we could find a way to only run tests for the scrapers that were changed), but that's very low prior.

PR #427, opened 2022-04-21
Remove Airflow infrastructure

smcclure17:
There's a lot more cleanup that could be done, especially in `services/`, but I'm not 100% sure what pieces are still needed at the moment


mikelehen (approved):
Thanks for the cleanup!

PR #429, opened 2022-05-23
Remove NYT `live/` files, fix VA scraper

smcclure17:
NYT is updating the yearly county files and the standard state/us files, so we no longer need to grab data from the `live/` files (see https://github.com/nytimes/covid-19-data/commit/5552a312c25189ff92a32dd15c32304a81c0e316)


mikelehen (approved):
Thanks for fixing / cleaning up! 🎉

PR #431, opened 2022-05-26
Tweaks for Prefect Cloud, etc.

mikelehen:
Updates requirements and services based on setting up Prefect Cloud.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,3 +20,5 @@ jmespath==0.10.0
 textract
 html5lib==1.1
 click==8.0.2
+fastparquet==0.8.1
+prefect==1.2.2
--- a/services/prefect/prefect-cloud-agent.service
+++ b/services/prefect/prefect-cloud-agent.service
@@ -1,17 +1,18 @@
 [Unit]
-Description=prefect agents
+Description=Prefect Cloud Agent
 
 [Service]
 Environment="DATAPATH=/home/sglyon/scraper-outputs"
 Environment="COVID_DB_CONN_URI=<postgress-url>"
+Environment="PREFECT__CLOUD__API_KEY=<prefect-api-key>"
+
 User=sglyon
 Group=sglyon
 Type=simple
 WorkingDirectory=/home/sglyon/can-scrapers
-ExecStart=/home/sglyon/miniconda/envs/prefect-can-scrapers/bin/prefect agent local start --api https://can-prefect.valorumdata.com/graphql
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -20,3 +20,5 @@ jmespath==0.10.0
 textract
 html5lib==1.1
 click==8.0.2
+fastparquet==0.8.1
+prefect==1.2.2
```
mikelehen, line 24: After I reinitialized the conda environment, one of the tasks was failing due to missing fastparquet.  And as you know prefect wasn't included (for some reason 🤷 ).

services/prefect/prefect-agent.service
```diff
@@ -3,6 +3,7 @@ Description=prefect agents
 
 [Service]
 Environment="DATAPATH=/home/sglyon/scraper-outputs"
+Environment="COVID_DB_CONN_URI=<postgress-url>"
```
mikelehen, line 6: This was on the system but not in github, so I'm adding it...  though perhaps we can just delete this file if Prefect Cloud works out. 🤷

requirements.txt
```diff
@@ -7,7 +7,7 @@ pandas==1.3.4
 psycopg2-binary==2.8.6
 pyppeteer==0.2.2
 pytest==6.1.2
-requests==2.24.0
+requests==2.27.1
```
mikelehen, line 10: I had to update this to avoid a urllib3 dependency confilct between requests and prefect.

requirements.txt
```diff
@@ -20,3 +20,5 @@ jmespath==0.10.0
 textract
 html5lib==1.1
 click==8.0.2
+pyarrow==8.0.0
```
mikelehen, line 23: After reinitializing the conda environment, some flows failed because pandas needs `fastparquet` or `pyarrow` for parquet support.  I actually tried `fastparquet` first and it seemed to work, except the UpdateParquetFiles task generated corrupt parquet files. 😬 Switching to `pyarrow` seems to work though.

requirements.txt
```diff
@@ -20,3 +20,5 @@ jmespath==0.10.0
...
+prefect==1.2.2
```
mikelehen, line 24: Not sure why this wasn't included before. 🤷

smcclure17: It looks like originally these were installed during the VM setup, https://github.com/covid-projections/can-scrapers/blob/main/services/prefect/setup_gcp_instance.sh#L57 but I think it definitely makes more sense to have them here in the requirements

smcclure17 (approved):
looks good!

thanks for tracking all of this down 🙌 🙌

PR #437, opened 2022-08-02
Add CDC Monkeypox scraper

smcclure17:
Add a scraper to pull in the cumulative Monkeypox cases reported by the CDC. 

This ingests the Monkeypox data in the existing Covid schema/infrastructure without any distinction (beyond the variable name) which is something I think we'll want to clean up if we decide to further invest in Monkeypox data (see my comments below) but until then I'm not sure it's worth the effort.

### DIFF
--- a/can_tools/bootstrap_data/covid_categories.csv
+++ b/can_tools/bootstrap_data/covid_categories.csv
@@ -1,5 +1,6 @@
 group,category
 case,cases
+case,monkeypox_cases
 case,cases_probable
 case,cases_confirmed
 case,b117_cases
--- a/can_tools/__init__.py
+++ b/can_tools/__init__.py
@@ -38,13 +38,45 @@ def all_subclasses(cls):
     )
 
 
-def scrapers_for_flow(cls):
-
+def unblocked_scrapers(cls):
+    # TableauDashboard, CDCTestingBase and PennsylvaniaVaccineDemographics
+    # are base classes that are not functional scrapers, so we explicitly remove them.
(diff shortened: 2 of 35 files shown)


can_tools/bootstrap_data/covid_categories.csv
```diff
@@ -1,5 +1,6 @@
 group,category
 case,cases
+case,monkeypox_cases
```
smcclure17, line 3: This does stray us a bit away from the intention of this schema and I'm not a huge fan of doing it this way, but: 

Ideally, I think we'd want to spin up a new schema that's parallel to the `covid` schema for this monkeypox data (which would consist of creating a `monkeypox_observations` table and a `monkeypox_us` view in Postgres, or something similar), but I'm not sure that's worth the investment at the moment.

smcclure17: Those schemas are created here: https://github.com/covid-projections/can-scrapers/blob/main/can_tools/models.py#L219

mikelehen: Yeah, I think this is the right approach for now.  If we get serious about collecting monkeypox data (e.g. scraping state dashboards and/or related monkeypox metrics), then we should revisit.

mikelehen (approved):
LGTM!  Do we need to add this to the `ACTIVE_SCRAPERS` list?

smcclure17:
> Do we need to add this to the ACTIVE_SCRAPERS list?

Ah, right 🤦 thanks!

PR #438, opened 2022-08-02
Remove validation task run from scraper flows

smcclure17:
Remove the validate task from scraper flows as validation is currently disabled (and has been for > a year). This leaves the actual/underlying validation code untouched

### DIFF
--- a/services/prefect/flows/generated_flows.py
+++ b/services/prefect/flows/generated_flows.py
@@ -123,13 +123,11 @@ def create_flow_for_scraper(ix: int, cls: Type[DatasetBase], schedule=True):
             d = create_scraper(cls)
             fetched = fetch(d)
             normalized = normalize(d)
-            validated = validate(d)
             done = put(d, connstr)
 
             d.set_upstream(sentry_sdk_task)
             normalized.set_upstream(fetched)
-            validated.set_upstream(normalized)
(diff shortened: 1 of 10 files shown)


services/prefect/flows/generated_flows.py
```diff
@@ -123,13 +123,11 @@ def create_flow_for_scraper(ix: int, cls: Type[DatasetBase], schedule=True):
             d = create_scraper(cls)
             fetched = fetch(d)
             normalized = normalize(d)
-            validated = validate(d)
             done = put(d, connstr)
 
             d.set_upstream(sentry_sdk_task)
             normalized.set_upstream(fetched)
-            validated.set_upstream(normalized)
-            done.set_upstream(validated)
```
mikelehen, line 132: I feel like we should probably actually remove the `validate()`, `_validate()` and `DatasetBase.validate()` functions or else keep these lines but comment them out with a comment that we haven't fully implemented the validate logic and we're disabling it since it costs us money.

mikelehen (approved):

services/prefect/flows/generated_flows.py
smcclure17: Yeah, fair point--I've gone ahead and removed the validate functions, and references to these steps in the docs 🙏

PR #439, opened 2022-08-08
Update CDC Monkeypox cases scraper

smcclure17:
The endpoint that's used to generate the [Monkeypox cases page](https://www.cdc.gov/poxvirus/monkeypox/response/2022/us-map.html) seems to be changing/not be super stable at the moment, causing our data to not always be up to date. This updates the scraper to pull directly from the HTML table on the page, instead of trying to hit the underlying endpoint.  

Two examples of the endpoints:
https://www.cdc.gov/poxvirus/monkeypox/response/modules/MX-response-case-count-US.json
https://www.cdc.gov/poxvirus/monkeypox/modules/data-viz/mpx-maps.json


mikelehen (approved):
Huh, lame that the CSV isn't stable but LGTM.

PR #443, opened 2022-10-31
Update CDC Testing scrapers

smcclure17:
Updates the CDC testing scrapers to the new weekly datasets. Disables the "Historical Changes" dataset for now, as there's no data in the dataset (causing the tests/scraper to fail). 

Currently, in `covid-data-model` we combine the data from the two datasets because The "As Originally Posted" data was typically a couple of days ahead of the "Historical Changes" data, but this might not be necessary going forward, depending on the content of the new weekly "Historical Changes" data


mikelehen (approved):
Looks good!

PR #444, opened 2022-10-31
Remove Direct Covid-Data-Tracker Scraper

BrettBoval:
It's the final scraper that directly calls covid.cdc.gov instead of data.cdc.gov.


smcclure17 (approved):
looks good to me! thanks for this

PR #447, opened 2022-12-12
Remove CDC case data

smcclure17:
We do not and would not be able to ingest this data into `covid-data-model` without a bit of work, and it adds a lot of data into the parquet file (9,212,545 rows) so I think it might be best to remove it until we potentially need it down the road


can_tools/scrapers/official/federal/CDC/cdc_variant_tracker.py
```diff
@@ -1,66 +0,0 @@
-import pandas as pd
```
smcclure17, line 1: Removing this as it's unused

PR #448, opened 2022-12-13
Use `dtypes` when loading db csv to reduce memory usage

smcclure17:
For a 1,000,000 row slice of the database view this reduced the loaded dataframe size from 128MB to 55MB. Hopefully this will also decrease the amount of memory needed for the parquet file updater. 

E.g.:
image


mikelehen (approved):
Nice!

PR #462, opened 2023-02-13
Deactivate vaccine demographic data scrapers

smcclure17:
Per our API docs we will be sunsetting demographic data after [2/15/2023](https://apidocs.covidactnow.org/updates#sunsetting-demographic-vaccine-data). As such, these scrapers will no longer be needed. 

A cleanup/follow-up chore would be to delete all the unused scrapers, but this current change will prevent the scrapers from being picked up by Prefect.

### DIFF
--- a/can_tools/__init__.py
+++ b/can_tools/__init__.py
@@ -80,8 +80,4 @@ ACTIVE_SCRAPERS = [
     scrapers.HawaiiVaccineCounty,
     scrapers.NCVaccineCounty,
     scrapers.NCVaccineState,
-    # Demographic data scrapers
-    *[scraper for scraper in ALL_SCRAPERS if scraper.demographic_data == True],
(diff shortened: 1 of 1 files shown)


can_tools/__init__.py
```diff
@@ -80,8 +80,4 @@ def unblocked_scrapers(cls):
     scrapers.HawaiiVaccineCounty,
     scrapers.NCVaccineCounty,
     scrapers.NCVaccineState,
-    # Demographic data scrapers
-    *[scraper for scraper in ALL_SCRAPERS if scraper.demographic_data == True],
-    # Monkeypox scrapers
-    scrapers.CDCCurrentMonkeypoxCases,
```
smcclure17, line 86: drive-by removal as we don't use the monkeypox data for anything

PR #465, opened 2023-03-21
Update caching to allow for non-etag checking

smcclure17:


can_tools/scrapers/official/base.py
```diff
@@ -1222,20 +1223,55 @@ def get_dataset(self, body, url) -> str:
         return rawJson
 
 
-class ETagCacheMixin:
+class CacheMixin(ABC):
```
smcclure17, line 1226: Create an abstract class that will be used by different caching implementations (for now just the old Etag caching, and the system used for the CDC scraper)

Issue #468, opened 2023-03-29
Migrate from Prefect Cloud 1 --> Prefect Cloud 2 by July 15th, 2023


smcclure17:
Context: https://www.prefect.io/guide/blog/freezing-legacy-prefect-cloud-1-accounts-on-starter-and-standard-plans/

We use Prefect Cloud 1 at the moment. I believe this is a separate task from updating our workflows from Prefect 1 to Prefect 2, which requires significant code changes

smcclure17:
Need to make a go/no-go decision on whether or not to continue running scrapers? Probably contingent on data landscape after [May 11th](https://oig.hhs.gov/coronavirus/covid-flex-expiration.asp)

PR #469, opened 2023-06-10
Upgrade dependencies in order to use `Prefect v2`, use `pipenv` as package manager, rip out unused scrapers

smcclure17:
I got into a little bit of dependency hell working on #468. Using pipenv should simplify our environment handling. 

Removing unused/old scrapers helped trim some dependencies that were causing conflicts. 

## TODO:

- [x] Specify package versions in `Pipfile`
- [ ] Update docs (document pipenv)
- [x] Make go/no-go decision on whether to fix failing scrapers (Philadelphia and Vermont)


.github/workflows/scraper-ci.yml
```diff
@@ -17,17 +17,19 @@ env:
   CENSUS_API_KEY: ${{ secrets.CENSUS_API_KEY }}
```
smcclure17, line 17: Update CI/CD workflow to use pipenv instead of installing the old `requirements.txt` directly

Pipfile
```diff
@@ -4,33 +4,33 @@ verify_ssl = true
 name = "pypi"
```
smcclure17, line 4: I guess we already had a pipenv/pipfile, but I don't think we typically used it 🤷 

This updates the dependencies to the current versions and adds some scripts for testing

can_tools/scrapers/__init__.py
```diff
@@ -76,90 +45,10 @@
     HHSReportedPatientImpactHospitalCapacityState,
```
smcclure17, line 45: Remove unused/deleted scraper imports

requirements.txt
```diff
@@ -1,24 +0,0 @@
-BeautifulSoup4==4.9.3
```
smcclure17, line 1: This is no longer needed with the addition/creation of the `Pipfile`

can_tools/scrapers/official/federal/HHS/hhs_state.py
```diff
@@ -108,25 +106,18 @@ def normalize(self, data: str) -> pd.DataFrame:
...
 
         # Put into long form
-        out = df.melt(id_vars=["dt", "location"], value_vars=crename.keys())
+        out: pd.DataFrame = df.melt(
+            id_vars=["dt", "location"], value_vars=crename.keys()
+        )
         out.loc[:, "value"] = pd.to_numeric(
             out["value"].astype(str).str.replace(",", "").replace("nan", None)
         )
+        # Dropping missing values. As of 2023-06-11 all missing values are from
```
smcclure17, line 115: Scraper was failing tests b/c there were some NaN values from 2020. Not entirely sure what changed to cause this failure, but this fixes the scraper

PR #470, opened 2023-06-16
Rewrite scraper flows using Prefect v2

smcclure17:
#468


can_tools/db_util.py
```diff
@@ -28,18 +29,15 @@ def fast_append_to_sql(
         raise ValueError(msg)
 
     with closing(engine.connect()) as con:
-
         if engine.dialect.name == "postgresql":
-            dest = (
```
smcclure17, line 33: changes necessary due to bumping psycopg to version 2.9x

services/prefect/deployment.py
```diff
@@ -0,0 +1,43 @@
...
+
+@main.command()
+@click.option(
+    "--choice",
+    "-c",
+    type=click.Choice(
+        ["all", "scraper", "clean-sql", "update-parquet", "github-pipeline-trigger"]
+    ),
+)
+def deploy_flows(choice: str):
```
smcclure17, line 27: allows us to deploy flows from the command line with `python -m services.prefect.deployment deploy-flows -c all`

services/prefect/flows/monitor_failing_scrapers.py
```diff
@@ -1,216 +0,0 @@
...
-from typing import Optional, List
-import datetime
-import dataclasses
-
-import requests
-import prefect
-from prefect import Flow, task
-from prefect.schedules import CronSchedule
-from prefect.tasks.secrets import PrefectSecret
-
```
smcclure17, line 11: This interacts with the old PrefectV1 graphQL API which has been replaced by a REST API

services/prefect/flows/utils.py
```diff
@@ -1,73 +0,0 @@
-import prefect
-from prefect.client import Client
-from prefect.engine.state import Skipped, State
-from prefect import Flow
-from typing import Optional, Set
-
-
-def skip_if_running_handler(obj, old_state, new_state):
```
smcclure17, line 8: No longer using these state handlers for simplicities sake. This means we won't be able to immediately tell when a flow has been "skipped" without looking at the task runs, which I think is fine

PR #471, opened 2023-06-27
Update Prefect services and documentation

smcclure17:


services/prefect/Makefile
```diff
@@ -1,18 +1,14 @@
 sync_services:
-	sudo mkdir -p /usr/lib/systemd/system/
```
smcclure17, line 2: Since we're using Prefect Cloud, we don't need our own server instance

